PR #2, opened 2020-08-28
dags: add load tasks

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR adds the tasks to load a local file into Google Cloud Storage.

### Why
We want to load files into Google Cloud Storage so that they can be moved into BigQuery more easily. This intermediate task handles the loading into GCS.

### Known limitations
Right now, only the history archives dag has the load tasks. This is because we will need to separate ledger entries changes  that are outputted by the ETL. Until then, the data should not be loaded into GCS.

### DIFF
--- a/dags/stellar_etl_airflow/build_load_task.py
+++ b/dags/stellar_etl_airflow/build_load_task.py
@@ -1,13 +1,19 @@
-from airflow.operators.python_operator import PythonOperator
 from google.cloud import bigquery
-from airflow.models import Variable
-from airflow import AirflowException
+from google.oauth2 import service_account
 
-def append_to_bigquery(file_path, table_id):
-    client = bigquery.Client()
-    table_id = Variable.get('project_name') + Variable.get('database_name') + table_id
+from airflow import AirflowException
+from airflow.models import Variable
+from airflow.operators.python_operator import PythonOperator
+
+def append_to_bigquery(filename, table_id, **kwargs):
+    key_path = Variable.get('api_key_path')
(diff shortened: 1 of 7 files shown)


dags/stellar_etl_airflow/build_load_task.py
```diff
@@ -0,0 +1,86 @@
...
+    
+    Parameter:
+        local_filepath - path to the local file to be uploaded 
+        gcs_filepath - path for the file in gcs
+        bucket_name - name of the bucket to upload to
+        mime_type - type of media to upload
+    Returns:
+        True if the upload is successful. Raises an Airflow error otherwise
+    '''
+    storage_service = build_storage_credentials()
```
debnil, line 35: Probably should validate some parameters before passing to this function, to guarantee safe usage. It can also double as documentation for future maintainers.

dags/stellar_etl_airflow/build_load_task.py
```diff
@@ -0,0 +1,86 @@
...
+def upload_to_gcs(filename, data_type, **kwargs):
+    '''
+    Uploads a local file to Google Cloud Storage and deletes the local file if the upload is sucessful.
+
+    
+    Parameter:
+        filename - name of the file to be uploaded 
+        data_type - type of the data being uploaded (transaction, ledger, etc)
+    Returns:
+        N/A
```
debnil, line 67: Looks like this returns `gcs_filepath`.

PR #15, opened 2020-09-02
dags: make exported filenames unique

### LINKED ISSUE #13: [3] Change exported file names to be unique for each batch
### What
We need to change the names of the files created by the export functions. File names should include the start and end ledger sequence numbers and the data type.

### Why
Airflow can run multiple DAGs in parallel. Right now, the file names for a given data type are the same, which leads to files possibly being overwritten or deleted.

### LINKED ISSUE #14: [1] Ensure exported ranges don't overlap
### What
Subtract 1 from the end ledger sequence number before exporting to remove overlap.

### Why
Right now, the start of one exported batch is the same ledger as the previous batch's end. This means we have duplicated information.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR changes the filenames of exported tasks. Now each one includes the start and end ledger. Additionally, there is no longer overlap between batches. Closes #13. Closes #14.

### Why
When tasks exported to files with the same name, information was overwritten or deleted prematurely. By ensuring each task creates and cleans up its own unique file, we don't have to worry about multiple DAG runs interfering with each other.

### Known limitations

### DIFF
--- a/dags/stellar_etl_airflow/build_export_task.py
+++ b/dags/stellar_etl_airflow/build_export_task.py
@@ -5,6 +5,7 @@ to be added to the PATH env variable.
 '''
 
 import json
+import logging
 
 from subprocess import Popen, PIPE
 
@@ -44,12 +45,12 @@ def execute_cmd(args):
     stdout, stderr = process.communicate()
     if process.returncode:
         raise AirflowException("Bash command failed", process.returncode, stderr)
-    return stdout, stderr
+    #return stdout, stderr
 
 def get_path_variables():
     return Variable.get('output_path'), Variable.get('core_exec_path'), Variable.get('core_cfg_path')
 
-def run_etl_cmd(command, filename, cmd_type, **kwargs):
+def run_etl_cmd(command, base_filename, cmd_type, **kwargs):
     '''
     Runs the provided stellar-etl command with arguments that are appropriate for the command type.
     The supported command types are: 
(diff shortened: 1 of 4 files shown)


dags/stellar_etl_airflow/build_export_task.py
```diff
@@ -62,28 +62,30 @@ def run_etl_cmd(command, filename, cmd_type, **kwargs):
...
     Returns:
-        output of the command, error
+        name of the file that contains the exported data
     '''
 
     start_ledger, end_ledger = parse_ledger_range(kwargs)
     output_path, core_exec, core_cfg = get_path_variables()
-    cmd_args = ['stellar-etl', command, '-o', output_path + filename]
+
+    batch_filename = '-'.join([start_ledger, end_ledger, base_filename])
```
debnil, line 74: The docstring indicates that the ledger range is appended to the filename. This logic pre-pends the ledger range to the filename.

dags/stellar_etl_airflow/build_load_task.py
```diff
@@ -80,18 +80,20 @@ def upload_to_gcs(filename, data_type, **kwargs):
         the full filepath in Google Cloud Storage of the uploaded file
     '''
 
-    start_ledger, end_ledger = parse_ledger_range(kwargs)
-    gcs_filepath = 'exported/' + data_type + '/' + '-'.join([start_ledger, end_ledger, filename])
+    filename = kwargs['task_instance'].xcom_pull(task_ids='export_' + data_type + '_task')
+    gcs_filepath = 'exported/' + data_type + '/' + filename
```
debnil, line 84: Nit: It's best practice to use [f-strings](https://realpython.com/python-f-strings/) to format strings in Python 3.x.

dags/stellar_etl_airflow/build_export_task.py
```diff
@@ -26,7 +27,7 @@ def parse_ledger_range(context):
     range_string = context['task_instance'].xcom_pull(task_ids='get_ledger_range_from_times')
     range_parsed = json.loads(range_string)
     start = range_parsed['start']
-    end = range_parsed['end']
+    end = max(range_parsed['end'] - 1, start)
```
debnil, line 30: What's the use case of this `max` logic, rather than reading the "end" entry from the parsed range?

dags/stellar_etl_airflow/default.py
```diff
@@ -8,5 +8,5 @@ def get_default_dag_args():
     'depends_on_past': False,
     'start_date': "2015-09-30T16:46:54+00:00",
     'retries': 5,
-    'retry_delay': timedelta(minutes=5),
+    'retry_delay': timedelta(seconds=30),
```
debnil, line 11: Why this retry delay?

Side note: in future, small unrelated changes like this should likely be pulled out into a separate PR. (Unless, of course, there's a relevance that I'm missing.)

dags/stellar_etl_airflow/build_export_task.py
Isaiah-Turner: In order to make sure that each file has a unique range of information, we need to subtract 1 from the end of the range. Otherwise, the end of one range is the same as the start of the next. The max is there because in some cases a file might only include a single ledger, so subtracting would result in a range where `end < start`.

dags/stellar_etl_airflow/default.py
Isaiah-Turner: Oh, this was a change I made for testing. I'll revert it.

PR #20, opened 2020-09-09
dags: added merge task

### LINKED ISSUE #9: [5] Add tasks to handle updates/deletes in BigQuery
### What
We need to Airflow tasks that allow for updating or deleting rows from a BigQuery table based on a file in GCS storage.
### Why
For ledger entry changes (accounts, offers, and trustlines), we store a point in time representation. Since this point in time state is mutable, we need to apply changes to it as the ledger progresses.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR adds a new Airflow task that merges a GCS file and a BigQuery table. Closes #9.

### Why
This task is required for data types like accounts, offers, and trustlines where we want to store point in time state. We can't perform a blind append for these tables. Instead, we need to update them with the changes we receive.


dags/stellar_etl_airflow/build_merge_table_task.py
```diff
@@ -0,0 +1,131 @@
...
+from airflow.contrib.hooks.gcs_hook import GoogleCloudStorageHook
+
+from google.cloud import bigquery
+from google.oauth2 import service_account
+
+
+
+def read_gcs_schema(data_type):
+    '''
+    Reads a the schema file corresponding to data_type from Google Cloud Storage and parses it.
```
debnil, line 19: Nit: "a the" -> the. 

Separately, how do we know this file exists? We likely should check if this file exists, and generate it if it doesn't.

dags/stellar_etl_airflow/build_merge_table_task.py
```diff
@@ -0,0 +1,131 @@
...
+    Generates the SQL queries that will insert or update a row in a table based on the provided schema.
+    
+    Parameters:
+        schema - an array of dictionaries, each containing the name, type, and description of a schema field 
+        source_table_alias - the name of the table being used as a data source
+    Returns:
+        the insert query and update query
+    '''
+
+    insert = []
```
debnil, line 45: Small preference for doing this through list comprehension, rather than loop, due to performance time. For example:
```
insert = [field['name'] for field in schema]
update = [f'{field['name']} = {source_table_alias}.{field['name']} for field in schema]
```

dags/stellar_etl_airflow/build_merge_table_task.py
```diff
@@ -0,0 +1,131 @@
...
+    update = []
+    for field in schema:
+        name = field['name']
+        insert.append(name)
+        update.append(f'{name} = {source_table_alias}.{name}')
+    insert_list = ', '.join(insert)
```
debnil, line 51: For even better performance, we can even do the join and list comprehension in one line. I feel less strongly about this, because there is a readability impact here. But it might be instructive to time both of those and see if there is a significant savings cost for the size of data you're handling.

dags/stellar_etl_airflow/build_merge_table_task.py
```diff
@@ -0,0 +1,131 @@
...
+        the parsed schema
+    '''
+
+    gcs_hook = GoogleCloudStorageHook(google_cloud_storage_conn_id='google_cloud_platform_connection')
+    schema_fields = json.loads(gcs_hook.download(
+        Variable.get('gcs_bucket_name'),
+        f'schemas/{data_type}_schema.json').decode("utf-8"))
+    return schema_fields
+
+def generate_queries_from_schema(schema, source_table_alias):
```
debnil, line 34: Should we split this into two functions, rather than just one? The performance benefit does justify a single function. But given that this function doesn't make any network requests, we can better optimize performance through internal methods like list comprehensions (example below). Also, having two functions would be a more readable and understandable function signature. It's weird to expect a tuple - just a single query feels more intuitive.

dags/stellar_etl_airflow/build_merge_table_task.py
```diff
@@ -0,0 +1,131 @@
...
+    Returns:
+        the newly created task
+    '''
+
+    dataset_name = Variable.get('bq_dataset')
+    true_table_id = Variable.get('table_ids')[data_type]
+    dest_alias = 'T'
+    source_alias = 'S'
+    
+    equality_comparison = ''
```
debnil, line 72: Preference for dictionary over if-else as switch-case logic in Python. Here's an [example](https://jaxenter.com/implement-switch-case-statement-python-138315.html). Also likely worth decomposing this into a separate function called `generate_equality_comparison` for readability.

dags/stellar_etl_airflow/build_merge_table_task.py
```diff
@@ -0,0 +1,131 @@
...
+    '''
+
+    key_path = Variable.get('api_key_path')
+    credentials = service_account.Credentials.from_service_account_file(key_path)
+    client = bigquery.Client(credentials=credentials, project=credentials.project_id)
+
+    gcs_filename = kwargs['task_instance'].xcom_pull(task_ids=f'load_{data_type}_to_gcs')
+    schema = read_gcs_schema(data_type)
+    external_config = bigquery.ExternalConfig('NEWLINE_DELIMITED_JSON')
+    external_config.source_uris = [f'gs://{gcs_filename}']
```
debnil, line 115: Do we need to check that this external file exists in GCS? And more importantly, do we need to do any cleanup after uploading?

I think it's fine to just load for now (without deletes), so we have cloud documents in GCS. But that should be noted somewhere, so a future user doesn't get surprised by a massive cloud storage folder.

debnil (approved):
Looks great overall! I've requested some stylistic changes and error handling, pre-merge.

PR #21, opened 2020-09-09
dags: fix stellar-core output folder

### LINKED ISSUE #19: [2] Change unbounded export to send output to single folder
### What
Right now, the unbounded core exports data using the same logic as the other exports. This means it exports to a folder with a name like 31518143-31518143-changes.txt. The numbers change based on the time when the DAG is run. We want to the behavior so that it exports to a folder defined by an Airflow variable.
### Why
Having a folder with a .txt label is incredibly confusing. Additionally, the file sensors that export to BigQuery look for data in a folder defined by an Airflow variable. Without ensuring that data is exported there as well, the ETL breaks down.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR ensures that the stellar core instances write changes to the correct folder, which is determined by an airflow variable. Closes #19.

### Why
The file sensors that export to BigQuery look for data in a folder defined by an Airflow variable. Without ensuring that data is exported there as well, the ETL breaks down.


dags/stellar_etl_airflow/build_export_task.py
```diff
@@ -72,16 +72,16 @@ def run_etl_cmd(command, base_filename, cmd_type, **kwargs):
     output_path, core_exec, core_cfg = get_path_variables()
 
     batch_filename = '-'.join([start_ledger, end_ledger, base_filename])
-    cmd_args = ['stellar-etl', command, '-o', output_path + batch_filename]
+    cmd_args = ['stellar-etl', command]
 
     if cmd_type == 'archive':
-        cmd_args.extend(['-s', start_ledger, '-e', end_ledger])
+        cmd_args.extend(['-s', start_ledger, '-e', end_ledger, '-o', output_path + batch_filename])
```
debnil, line 78: Since `output_path + batch_filename` is used four times, it's a good idea to just define this as a separate variable to avoid any issues.

dags/stellar_etl_airflow/build_load_task.py
```diff
@@ -80,8 +80,8 @@ def upload_to_gcs(data_type, prev_task_id, **kwargs):
         the full filepath in Google Cloud Storage of the uploaded file
     '''
 
-    filename = kwargs['task_instance'].xcom_pull(task_ids='export_' + data_type + '_task')
-    gcs_filepath = f'exported/{data_type}/{filename}'
+    filename = kwargs['task_instance'].xcom_pull(task_ids=prev_task_id)
+    gcs_filepath = f'exported/{data_type}/{os.path.basename(filename)}'
```
debnil, line 84: Did not know you could call a function within an f-string. Cool!

dags/stellar_etl_airflow/merge_table_task.py
```diff
@@ -0,0 +1,88 @@
+import json
```
debnil, line 1: I'd restructure this file a bit. It looks like it has a bunch of duplicated code with `build_merge_table_task.py`. It's worth investing in a proper decomposition now, rather than having to do double the maintenance and fixing later.

debnil (approved):
LGTM! Some comments about structure and decomposition.

dags/stellar_etl_airflow/merge_table_task.py
Isaiah-Turner: Whoops, that wasn't supposed to be there.

PR #27, opened 2020-09-16
dags: added logging to  tasks

### LINKED ISSUE #4: [3] Investigate Airflow logging
### What
We want to figure out where Airflow sends logs, how to view them, and how to add more information or remove info by filtering.

### Why
Airflow logging may not be enough for Stellar's use case. If we want to extend the existing logging system, we want to make sure that we understand the current system.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR adds more logging to Airflow tasks. This logging details things like where tasks are in their process, what files they interact with, and BigQuery job results. Closes #4 .

### Why
We want more detailed logging in order to make future debugging easier. It also lets us see what is being modified and see more in-depth information on the results of tasks.

### Known limitations
From what I can tell, logging to the log files instead of the command line only works for functions called by PythonOperators. This means this PR doesn't add any additional logging to other operators. Fortunately, other operators tend to have their own logging or perform simple tasks that don't require detailed logs. If we need more logging for these other operators, we could extend the operators ourselves and add logging where desired.

### DIFF
--- a/dags/bucket_list_dag.py
+++ b/dags/bucket_list_dag.py
@@ -5,7 +5,6 @@ to stop exporting. This end ledger  is determined by when the Airflow DAG is run
 when initializing the tables in order to catch up to the current state in the network, but should not be scheduled to run constantly.
 '''
 import json
-import logging
 from stellar_etl_airflow.build_export_task import build_export_task
 from stellar_etl_airflow.build_time_task import build_time_task
 from stellar_etl_airflow.build_load_task import build_load_task
--- a/dags/bucket_list_dag.py
+++ b/dags/bucket_list_dag.py
@@ -5,7 +5,7 @@ to stop exporting. This end ledger  is determined by when the Airflow DAG is run
 when initializing the tables in order to catch up to the current state in the network, but should not be scheduled to run constantly.
 '''
 import json
-
+import logging
 from stellar_etl_airflow.build_export_task import build_export_task
 from stellar_etl_airflow.build_time_task import build_time_task
 from stellar_etl_airflow.build_load_task import build_load_task
@@ -24,6 +24,7 @@ dag = DAG(
 )
 
 file_names = Variable.get('output_file_names', deserialize_json=True)
+logger = logging.getLogger("airflow.task")
 
 time_task = build_time_task(dag, use_next_exec_time=False)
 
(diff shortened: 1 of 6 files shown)


debnil (changes requested):
Instead of passing around the logger as a parameter, the best practice in Python is to instantiate a per-module logger. This [Bloomberg blogpost](https://www.techatbloomberg.com/blog/guide-logging-python/) has some helpful code examples, and this [StackOverflow post](https://stackoverflow.com/questions/5974273/python-avoid-passing-logger-reference-between-functions) has more conversation about various Pythonisms.

I will say that this approach will be substantially simpler in code. But you're already pretty far down a different path. It might be cleaner for you to scrap this PR and use it as a reference for a different one following the best practice.

I'm really sorry about that confusion - this is mostly on me for not looking into this earlier. It's also a good note for the future, when working in a newer codebase, just double check language idioms and best practices before implementation. (If the codebase has established examples, then default to those.)

dags/process_unbounded_core_dag.py
```diff
@@ -20,6 +20,7 @@
     schedule_interval='@once',
 )
 
+
```
debnil, line 23: Nit: remove newline.

dags/stellar_etl_airflow/build_apply_gcs_changes_to_bq_task.py
```diff
@@ -124,7 +128,7 @@ def apply_gcs_changes(data_type, **kwargs):
     Data types should be: 'accounts', 'offers', or 'trustlines'.
 
     Parameters:
-        data_type - type of the data being uploaded; should be string 
+        data_type - type of the data being uploaded; should be string
```
debnil, line 131: Nit: remove space.

dags/stellar_etl_airflow/build_export_task.py
```diff
@@ -73,7 +70,7 @@ def build_export_task(dag, cmd_type, command, filename):
 
     # echo the output file so it can be captured by xcom; have to run bash to combine commands
     full_cmd = f'bash -c "{etl_cmd} && echo {output_file}"'
-
+
```
debnil, line 73: Nit: remove tab.

dags/stellar_etl_airflow/build_load_task.py
```diff
@@ -103,7 +105,7 @@ def build_load_task(dag, data_type, prev_task_id):
     
     Parameters:
         dag - the parent dag
-        data_type - type of the data being uploaded (transaction, ledger, etc)
+        data_type - type of the data being uploaded; should be string
```
debnil, line 108: I'd keep the examples, and just add the type at the end - unless those examples are now invalid, in which case adding new examples is even more helpful!

dags/bucket_list_dag.py
```diff
@@ -5,7 +5,7 @@
 when initializing the tables in order to catch up to the current state in the network, but should not be scheduled to run constantly.
 '''
 import json
-
+import logging
```
debnil, line 8: Do you need the import if you don't invoke the logger in the file? (i.e., need to call at the top level because invoked modules are logging)

debnil (approved):
LGTM! Some remaining cleanup, nothing too big - impressive turnaround :)

PR #42, opened 2020-11-05
Add orderbook info and extensions to readme

Isaiah-Turner:
### What
This PR adds information about the new export order book DAGs. It also adds information about how to add new DAGs or extend existing ones. 

### Why
We need to keep the documentation up-to-date with the current features. We also want to have a guide for extensions for the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,10 +330,9 @@ dag = DAG(
 	schedule_interval=None,
 )
 ```
-Feel free to add more arguments or customize the existing ones. The documentation for a DAG is available [here](https://airflow.apache.org/docs/stable/_api/airflow/models/dag/index.html).
+The get_default_dag_args() is defined in the `dags/stellar-etl-airflow/default.py` file.
 
-### Adding tasks to existing DAGs
-If you have created a new DAG, or wish to extend an existing DAG, you can add tasks to it by calling the various `create_X_task` functions that are in the repository. See [here](https://airflow.apache.org/docs/stable/concepts.html#relations-between-tasks) for details on how to create dependencies between tasks.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -308,6 +317,34 @@ The airflow_variables.txt file provides a set of default values for variables.
...
 
+## Extensions
+This section covers some possible extensions or further work that can be done.
+
+### Adding New DAGs
+Adding new DAGs is a fairly straightforward process. Create a new python file in the `dags` folder. Create your dag object using the code below: 
+```
+dag = DAG(
+	'dag_id',
+	default_args=get_default_dag_args(),
```
debnil, line 328: Where is the `get_default_dag_args()` function defined? Might be useful to point out where below - not a huge change, but marginally more useful when onboarding.

README.md
```diff
@@ -308,6 +317,34 @@ The airflow_variables.txt file provides a set of default values for variables.
...
+	schedule_interval=None,
+)
+```
+Feel free to add more arguments or customize the existing ones. The documentation for a DAG is available [here](https://airflow.apache.org/docs/stable/_api/airflow/models/dag/index.html).
+
+### Adding tasks to existing DAGs
+If you have created a new DAG, or wish to extend an existing DAG, you can add tasks to it by calling the various `create_X_task` functions that are in the repository. See [here](https://airflow.apache.org/docs/stable/concepts.html#relations-between-tasks) for details on how to create dependencies between tasks.
+
+### Adding New Tasks
+Adding new tasks is a more involved process. You likely need to add a new python file in the `dags/stellar_etl_airflow` folder. This file should include a function that creates and returns the new task, as well as any auxiliary functions related to the task.
```
debnil, line 339: Is there a particular task that would be a good reference - maybe one not too complicated? If so, it would be helpful to add that here.

debnil (approved):
LGTM! Couple small requests for additional details and examples, but overall structure looks good.

PR #45, opened 2020-11-20
Fix dag ids in trigger_next task

Isaiah-Turner:
When the orderbook export DAG was added, the IDs for the dags did not get changed in the trigger_next task.


debnil (approved):
LGTM!

PR #47, opened 2020-11-30
Switch to KubernetesPodOperator

### LINKED ISSUE #46: [5] Switch to KubernetesPodOperators
### What
Instead of using the DockerOperator to access the ETL, we need to use the KubernetesPodOperator.

### Why
The Docker operator requires that we have Docker installed on the host machine. In production, this means we need Docker on the airflow worker pods. We have to mount docker onto the airflow workers, which has security concerns. Instead, we can use the KubernetesPodOperator, which won't require Docker.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR switches the tasks that use the ETL from the DockerOperator to the KubernetesPodOperator. It also introduces new variables that make it possible to customize the Pod and its volumes. Closes #46.

### Why
The DockerOperator had security concerns. We would have to provide root privileges to the container so it could run Docker. This switch prevents those concerns.

### Known limitations
Need to document the newly added variables.

### DIFF
--- a/airflow_variables.txt
+++ b/airflow_variables.txt
@@ -7,6 +7,7 @@
     "gcs_exported_data_bucket_name": "stellar-testing-us",
     "image_name": "stellar/stellar-etl:latest",
     "image_output_path": "/etl/exported_data/",
+    "kube_config_location": "",
     "local_output_path": "/home/airflow/etlData/",
     "namespace": "default",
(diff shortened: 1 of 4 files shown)


debnil (approved):
Great work, LGTM!

PR #52, opened 2020-12-11
Airflow variable fixes

Isaiah-Turner:
This PR removes an unused variable. It also updates the time_to_ledger kubernetes pod to use the Airflow variable for namespace.


debnil (approved):
LGTM! Can you rename this PR to be a bit more expressive? Will make reading commit history a bit easier.

PR #56, opened 2020-12-16
Fix duplicate information in BigQuery

Isaiah-Turner:
### What
This PR changes the history archive export. Now, it uses merge tasks to ensure rows are only added if they are new.

### Why
There was duplicated data in the Airflow tables.


debnil (approved):
LGTM!

PR #58, opened 2020-12-17
Create branch with dags at top level

Isaiah-Turner:
### What
This PR creates a branch with the DAGs at the top level

### Why
For a git-sync container in Kubernetes, we need to have the DAGs at the top level of the repository so that they are visible to Airflow. If the DAGs are in an internal folder in the repository, Airflow won't recursively check for them.


debnil (commented):
LGTM!

PR #59, opened 2020-12-18
Add assets to history archive DAG

Isaiah-Turner:
This PR adds assets to the history archive DAG. It also updates documentation and variables to account for assets.


schemas/assets_schema.json
```diff
@@ -0,0 +1,22 @@
+[
+  {
+    "mode": "NULLABLE",
+    "name": "id",
+    "type": "FLOAT"
```
debnil, line 5: Has this data type been confirmed? Just making sure it shouldn't be a string.

debnil (approved):
LGTM! One small question.

schemas/assets_schema.json
Isaiah-Turner: The hash function returns a number. An example of an asset is `{"asset_code":"","asset_issuer":"","asset_type":"native","id":12638146518625398189}`. The numbers can be very large, so BigQuery uses a float to store them.

PR #60, opened 2020-12-21
Add xcom sidecar variable

Isaiah-Turner:
### What
This PR adds a variable that sets the image for the xcom sidecar that extracts information from the kubernetes pod operators.

### Why
DockerHub limited the pull rate of images. We need to point all the images to our private image repository. The sidecar image was hardcoded, so this PR turns it into a variable.


dags/stellar_etl_airflow/pod_generator.py
```diff
@@ -0,0 +1,598 @@
+"""
+This module provides an interface between the previous Pod
```
debnil, line 2: Is this module also adapted from an Airflow module in a recent version? If so, that file should be linked.

debnil (approved):
LGTM!

PR #61, opened 2020-12-22
Remove overlap ledger

Isaiah-Turner:
### What
This PR subtracts 1 from the end of the ledger range for archive export commands. 

### Why
This ensures there is no overlap between exports. If there is overlap, then if two overlapping files are uploaded concurrently, the table will have repeated rows.


debnil (approved):
Great catch - LGTM!

PR #62, opened 2020-12-22
Update offer schema

Isaiah-Turner:
Now that we have assets, the offer schema should use asset ids to represent the buying and selling assets. These ids are now floats instead of strings.


debnil (approved):
LGTM!

PR #64, opened 2020-12-22
Add image_pull_policy and resources variables

Isaiah-Turner:
This PR adds image_pull_policy and resources variables to allow for greater customization of K8 pods.


debnil (approved):
LGTM!

PR #65, opened 2020-12-22
Added trigger rule for process DAGs

Isaiah-Turner:
### What 
This PR changes the `trigger_next` task in the `process_unbounded_core_changes` and `process_unbounded_core_orderbooks` DAGs to always trigger. 

### Why
If a processing run fails and the `trigger_next` task doesn't run, then future batches will not be uploaded. The only way to fix this was to manually trigger another process DAG run. Now, the trigger will happen automatically.


debnil (approved):
LGTM! Nice change.

PR #66, opened 2020-12-23
Long running fixes

Isaiah-Turner:
### What
This PR switches the factEvent branch in the process orderbook dag to use a unique merge instead of a blind append. It also sets the retries for long running tasks to 0.

### Why
We need to use merges instead of appends to ensure tasks are idempotent. Setting retries to 0 prevents us from reingesting huge amounts of data if a failure occurs in a long running task. It does require some more manual work, which is detailed in the `Long-Running Task Failures` section of the [Handling Failures](https://docs.google.com/document/d/1IIpDW0nrR-RtXx0dZDAViCtgjn78hhczYKoj71I65EA/edit#heading=h.c1mbxnin3mci) document.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,10 @@ This repository contains the Airflow DAGs for the [Stellar ETL](https://github.c
 	  - [Normal Variables](#normal-variables)
 	  - [Kubernetes Specific Variables](#kubernetes-specific-variables) 
 - [Execution Procedures](#execution-procedures)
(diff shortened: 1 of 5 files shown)


debnil (approved):
LGTM!

Request: since the google doc isn't public, could you copy-paste any relevant parts of that that aren't public yet into the README? It can be either pre-merge or in a follow-up PR, your call.

PR #69, opened 2021-04-23
poststart.sh: check if file has size before moving

acharb:
**WHAT**

adding a check to see if the file on the local pod has size before moving it to the google cloud shared folder

**WHY**

The way DockerOperator in `build_export_task` works when outputting to a file is it creates the file first, then writes to it. The current poststart script is constantly checking for the presence of a file, and if found and unopened, it moves it. It seems it was moving it before the file was written to but after created, because it shows up in gcs empty, even though it originally had data.

This only shows up when the ledger range is larger. I believe because gathering the data takes longer, so it widens the window the poststart script has to move the file preemptively.

This solution is essentially:
- move the .txt file to gcs if it has data
- for the load task if there is no .txt file, then no data to upload so end successfully (if for whatever reason the export task fails, then the load task wouldn't start anyways bc it's downstream)

### DIFF
--- a/dags/stellar_etl_airflow/build_load_task.py
+++ b/dags/stellar_etl_airflow/build_load_task.py
@@ -46,6 +46,7 @@ def attempt_upload(local_filepath, gcs_filepath, bucket_name, mime_type='text/pl
     '''
 
     storage_service = build_storage_service()
+    logging.info('File size is: %s kb' % str(os.path.getsize(local_filepath) / 1000))
     if os.path.getsize(local_filepath) > 10 * 2 ** 20:
         media = MediaFileUpload(local_filepath, mime_type, resumable=True)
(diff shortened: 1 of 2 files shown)


nikhilsaraf (commented):
Makes sense. Can approve if this is in final stage

nikhilsaraf:
Would it make sense to add logging, if we will be able to view it somewhere. That way we can retroactively see the size of the file and what time the request came in to move it etc.

If it’s possible to add and view such logs in GCP production then we should consider adding it in.

acharb:
added a file size log here that should help with debugging: https://github.com/stellar/stellar-etl-airflow/pull/69/commits/53f042b685f38dff8b0e8b5277269790c2e24e80

nikhilsaraf (approved):
LGTM, thanks!

PR #72, opened 2021-08-31
Update schemas, Reference Airflow Operators

sydneynotthecity:
This PR covers 3 primary changes to the `stellar-etl-airflow` repo. These changes help with logging, reconciliation to Horizon and making sure that the repo can support the updates in `stellar-etl` Docker image. 

#### Docker and Kube Operators
The original code used custom written operators (found under `/dags/stellar_etl_airflow/docker_operator.py` and `/dags/stellar_etl_airflow/kubernetes_pod_operator.py`) instead of the operators that come standard with Airflow. I believe this was because Cloud composer did not support the correct version of Airflow, however they do now 😄 
**Changes**
* all import statement import the Airflow operator instead of custom
* Logging in `DockerOperator` is enable with `tty=true` so that developers can view Docker logs 

#### Poststart.sh 
This script would prematurely move large files. We only see this behavior when a file exceeds >500MB and assume the Go code is closing the file between large writes to "batch" the write. The old script would move an incomplete file which resulted in an incomplete load. Added an additional check that compares file size for two times, 10 seconds apart. If the size matches, move the file, otherwise wait until the size comparisons match to move.

#### History Archives DAG
Some configuration changes that impact both functionality and cosmetics of the DAG:
* Airflow variables table names are updated  to match Horizon tables (Helix table names should be considered one-off names)
* Schedule runs every 3 hours. will update to 5 min when history is caught up
* Start date is moved to a closer date for testing purposes
* A sleep timer is introduced after the `export_*` tasks. The `poststart.sh` script will sleep for 10 seconds before moving. If the `load_*` task starts too early, it is possible the file will not be moved yet.

#### Schema Updates
Schemas for the `history_operations` and `history_transactions` tables are updated to match Hubble 1.0 and Horizon. See [193](https://github.com/stellar/hubble/pull/193) and [188](https://github.com/stellar/hubble/pull/188) for full column list.

The `test-hubble` GCP project uses all updated code in its `history_archive_export` DAG (3 hour cadence).


poststart.sh
```diff
@@ -1,11 +1,15 @@
...
 move_file_if_closed() {
-    fileSize=$(wc -c < "$1")
+    preFileSize=$(wc -c < "$1")
+    sleep 10;
+    postFileSize=$(wc -c < "$1")
     lsof "$1";
     # if lsof returns an error code of 1, it means the file is not opened by any other processes, allowing us to move it safely
-    if [[ $? == 1 && $fileSize -gt 0 ]]; then
-        mv "$1" /home/airflow/gcs/data/;
+    if [[ $? == 1 && $preFileSize -gt 0]]; then
```
acharb, line 9: 💡 can we add a comment here describing what's happening? Took me a second to understand since not great with bash. Something like
`# if the file size does not change after a 10 second sleep then the write is finished and we can move`

upload_static_to_gcs.sh
```diff
@@ -10,5 +10,5 @@ if [ -z "${airflow_bucket}" ]; then
     exit 1
 fi
 
-gsutil cp -r dags/* gs://"${airflow_bucket}"/dags/
-gsutil cp -r schemas/* gs://"${airflow_bucket}"/dags/schemas/
+gsutil -m cp -r dags/* gs://"${airflow_bucket}"/dags/
+gsutil -m cp -r schemas/* gs://"${airflow_bucket}"/dags/schemas/
```
acharb, line 14: ❓ jw what's the `-m` doing? Makes copying the files over faster?

dags/history_archive_dag.py
```diff
@@ -32,12 +34,21 @@
 The export tasks call export commands on the Stellar ETL using the ledger range from the time task.
 The results of the comand are stored in a file. There is one task for each of the data types that 
 can be exported from the history archives.
+
+The DAG sleeps for 30 seconds after the export_task writes to the file to give the poststart.sh
+script time to copy the file over to the correct directory. If there is no sleep, the load task 
+starts prematurely and will not load data.
 '''
 ledger_export_task = build_export_task(dag, 'archive', 'export_ledgers', file_names['ledgers'])
+ledger_export_task.post_execute = lambda **x: time.sleep(30)
```
acharb, line 43: ❓  do you know how much the added sleeps slow down the dag? Eg. the sleeps added here add 2.5 min per dag run + the poststart script 10 second sleeps. iirc a full catchup with the 3 hour windows took ~1 week (with errors). I guess we can test how long a couple days takes after this merge and then extrapolate to see if it's manageable for a full catchup, I agree making sure we don't miss data is important. wdyt?

acharb (approved):
lgtm! had two questions and a 💡 , but nothing blocking so going to approve

README.md
```diff
@@ -51,21 +51,21 @@ Create a new Cloud Composer environment using the command below or the [UI](http
 ```bash
 gcloud composer environments create <environment_name> --location=<project_location> \
 --zone=<project_zone> --disk-size=100GB --machine-type=n1-standard-4 \
---node-count=3 --python-version=3 --image-version=composer-1.12.0-airflow-1.10.10 \
---oauth-scopes=['https://www.googleapis.com/auth/cloud-platform'] \
+--node-count=3 --python-version=3 --image-version=composer-<version>-airflow-<version> \
```
acharb, line 54: ❓ oh I missed this, does it not matter the airflow version to use?

dags/stellar_etl_airflow/build_export_task.py
```diff
@@ -87,8 +87,7 @@ def build_kubernetes_pod_exporter(dag, command, etl_cmd_string, output_file):
     '''
     from airflow.kubernetes.volume import Volume
     from airflow.kubernetes.volume_mount import VolumeMount
-    from stellar_etl_airflow.kubernetes_pod_operator import KubernetesPodOperator
-
+    from airflow.contrib.operators.kubernetes_pod_operator import KubernetesPodOperator
```
acharb, line 90: very glad we can import the airflow ones here now btw 🎉

upload_static_to_gcs.sh
sydneynotthecity: it multithreads the `copy` statement. otherwise each file is sequentially copied to GCP. It only makes this command marginally faster, but as our codebase grows it'll be more useful.

poststart.sh
sydneynotthecity: Added a comment, will include in the next PR

dags/history_archive_dag.py
sydneynotthecity: Since the export tasks run in parallel, it actually would increase run time somewhere between 30 sec - 2.5 min depending on when each export finishes. The export tasks can vary in runtime (I've seen between 45 seconds and 4 minutes) and the sleep for 30 would just be appended to each parallel execution. 

I've been doing some benchmarking and yesterday completed a test of 1 month data (took 15 hours but I had decreased concurrency to 5 instead of 15 since we had memory errors....) My plan is to add worker nodes during the catch up and configure in such a way that it takes maximum a week to run full history.

README.md
sydneynotthecity: the version we were on is going to be deprecated from GCP support soon. I'll add a comment that the user should choose a stable composer/airflow version that works for them. I did not want to promote a version that would be soon deprecated.

PR #73, opened 2021-09-08
Build Batch Stats Logging

sydneynotthecity:
The current ETL pipeline had no batch execution metadata, which makes validation and reconciliation of the data more difficult. By adding in metadata about the batch execution, we are able to configure more precise validation. This also leads to better understanding of our batch execution pipeline over time. This PR also proposes table partitioning for the large, high transaction volume tables. Partitioning will help with data management and query efficiency as the tables grow.

### Batch Stats
Batch Stats is an internal function that is useful to maintenance and reconciliation of datasets. We are starting with a simple strategy and will build more as the need arises. To start, each table will have 3 columns: 
1. `batch_id` - DAG run id that corresponds to a unique DAG execution
2. `batch_run_date` - The batch execution start date. This is _not_ when the DAG executes, but rather the time period (run date) that is executing. This is correlated to ledger_close time.
3. `batch_insert_ts` - The batch insert timestamp is the current time in UTC. This will indicate if the batch was rerun or loaded during historical loading

There will be a consolidated tables for more batch metadata, called the `history_archives_dag_runs`. This table will have batch_id, batch_run_date (join keys) with start_ledger_sequence/end_ledger_sequence and table name.  For reconciliation purposes, it is useful to know the ledger start and end boundaries to validate that all data during the time period was moved.

All code updates support writing the new columns to existing datasets and row insertion to the batch stats table.

### history_archives DAG updates
The biggest update is how the data is loaded from GCS to Bigquery. In the old DAG version, there is a complex `MERGE` statement generated that scans a table and inserts/updates individual records if there is a match in the current dataset. While this works for reruns, it is inefficient because it scans an entire table, updating records where there are changes. When loading data in batches, it is more efficient (and simpler to support) a pattern that deletes an entire batch before reloading. The DAG has been redesigned to delete records with the current batch_id (if they exist) and re-insert the batch of records.

The DAG flow includes new tasks: 
* `write_<table>_stats` - task will write a record in the `history_archives_dag_runs` which will include the batch id, batch execution date, start ledger, end ledger and table name. Future plans to include batch run information (like number of records processed, number of errors) which will make reporting on data health/pipeline health easier.
* `delete_old_<table>_task` - will search for a partition/batch id combo and delete all records in the batch if found
* `gcs_to_bq_task` - the `build_apply_gcs_changes_to_bq` has been swapped for the simpler `gcs_to_bq` task. Since any records have been deleted in the prior task, we can use the `GoogleCloudStorageToBigQueryOperator` to insert the GCS file as a dataset to BQ. 

### DDL helpers
A new folder was introduced, `/dags/ddls` that will contain ad hoc scripts that will aid in table maintenance over time. These scripts should be executed as needed when tables need to be created or updated.

Two scripts have been added : 
1. `create_partitioned_tables.sh` - script will create all history_archives tables from schema files. The `history_ledgers`, `history_transactions` and `history_transactions` tables are partitioned tables due to size. Table partitioning helps with data management and cuts down on query costs when the partition is included in the query.
2. `update_table_schema.sh` - script will issue an update command to adjust table schemas as they change.


sydneynotthecity:
@nikhilsaraf for your awareness

dags/stellar_etl_airflow/build_batch_stats.py
```diff
@@ -0,0 +1,31 @@
...
+    start_ledger = '{{ ti.xcom_pull(task_ids="get_ledger_range_from_times")["start"] }}'
+    end_ledger = '{{ ti.xcom_pull(task_ids="get_ledger_range_from_times")["end"] - 1 }}'
+    batch_id = '{{ run_id }}'
+    batch_run_date = '{{ prev_execution_date.to_datetime_string() }}'
+    batch_start = datetime.now().strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+    INSERT_ROWS_QUERY = (
+        f"INSERT {DATASET_ID}.history_archives_dag_runs VALUES "
+        f"('{batch_id}', '{batch_run_date}', {start_ledger}, {end_ledger}, '{table}', '{batch_start}');"
```
acharb, line 18: not sure if it matters if this ran correctly, but to be safe
```suggestion
        f"('{batch_id}', '{batch_run_date}', '{start_ledger}', '{end_ledger}', '{table}', '{batch_start}');"
```

dags/stellar_etl_airflow/build_gcs_to_bq_task.py
```diff
@@ -15,16 +15,23 @@ def build_gcs_to_bq_task(dag, data_type):
...
+        partition - bool if the table is partitioned
     Returns:
         the newly created task
     '''
     
     bucket_name = Variable.get('gcs_exported_data_bucket_name')
     project_name = Variable.get('bq_project')
-    dataset_name = Variable.get('bq_dataset')
+    # dataset_name = Variable.get('bq_dataset')
+    dataset_name = 'test_gcp_airflow_internal_partitioned'
```
acharb, line 26: 💭  I think we should leave these as environment variables so they can be changed easier later on

dags/stellar_etl_airflow/build_load_task.py
```diff
@@ -99,6 +124,13 @@ def upload_to_gcs(data_type, prev_task_id, **kwargs):
     logging.info(f'Attempting to upload local file at {local_filepath} to Google Cloud Storage path {gcs_filepath} in bucket {bucket_name}')
     # TODO: consider adding a sanity check to make sure the filename shouldn't exist (ie no data to upload)
     if os.path.exists(local_filepath):
+        if batch_stats:
+            logging.info('Adding batch stats metadata to file.')
+            batch_date = kwargs['prev_execution_date'].to_datetime_string()
+            append_batch_stats(local_filepath, kwargs['run_id'], batch_date)
+            # remove original file since batch_stats writes out a new .jsonl file
+            os.remove(local_filepath)
```
acharb, line 132: ❗ a couple lines down (145) we also have this line of removing the local file. Which I think will panic if the file doesn't exist. Do we need to remove here or can we just remove at that line?

acharb (commented):
looks great 👍  I think adding batching will be a great improvement. Left a couple comments, one I think is blocking

dags/stellar_etl_airflow/build_batch_stats.py
sydneynotthecity: Since the `start_ledger` and `end_ledger` are defined as `int64`, they actually have to be numeric value in the `INSERT` statement. If we put quotes around the values, similar to JSON, the SQL query would try and insert as strings.

dags/stellar_etl_airflow/build_gcs_to_bq_task.py
sydneynotthecity: good catch, will update

dags/stellar_etl_airflow/build_load_task.py
sydneynotthecity: there are actually _two_ files that get written by the `export_task` and `build_load_task.py`. One file is the original file exported from history archives (export_<table>.txt), and the other file has batch stats appended (export_<table>.jsonl). The `*.txt` file is removed once the `*.jsonl` file is written and then [this](https://github.com/stellar/stellar-etl-airflow/blob/szwise-build-batch-stats/dags/stellar_etl_airflow/build_load_task.py#L133) updates the `local_filepath` to the `*.jsonl` name and removes the file once transferred. If this too confusing, I can keep the filenames separate, I had just repointed the variable to limit the amount of code refactoring.

acharb: ah I see thx for explaining. Ya  I'd vote for giving each a unique name to be safe + readability 

going to approve now since a minor change

acharb (approved):
lgtm!

PR #75, opened 2021-11-02
Update table schemas

sydneynotthecity:
In the `stellar-etl` repo, some [breaking changes](https://github.com/stellar/stellar-etl/pull/137) were made to outputted, transformed fields so that data transformations matched the Horizon API. Because of this, the downstream Airflow schemas needed to be updated so that the DAGs could still load tables. The changes are intended to make the data more usable to downstream consumers.

#### Changes

**Accounts**
* `amount` fields in the accounts table were updated from `INTEGER` to `FLOAT`
* batch run fields were appended to the table to help with maintenance and support

**History_Operations**
* New fields are added to `details` JSON blob for operation types 22 and 23 (Deposit and Withdraw liquidity) 
* Missing fields added to `details.claimants` so that claimable balances parse correctly

**History_Trades**
* all `base` and `counter` fields are updated to `buying` and `selling` prefixes to decouple from Horizon 
* liquidity pool fields added

**Trustlines**
* `amount` fields are updated from `INTEGER` to `FLOAT` (values were originally in stroops) 
* all batch run fields added
* liquidity pool fields added

**Offers**
* all batch run fields added
* `amount` fields are updated from `INTEGER` to `FLOAT` (values were originally in stroops)

**Liquidity_Pools**
Brand new table for liquidity pool data. The schema matches Horizon db with additional information for assets.

### DIFF
--- a/schemas/accounts_schema.json
+++ b/schemas/accounts_schema.json
@@ -7,17 +7,17 @@
   {
     "mode": "NULLABLE",
     "name": "balance",
-    "type": "INTEGER"
+    "type": "FLOAT"
   },
   {
     "mode": "NULLABLE",
     "name": "buying_liabilities",
-    "type": "INTEGER"
+    "type": "FLOAT"
   },
   {
     "mode": "NULLABLE",
     "name": "selling_liabilities",
-    "type": "INTEGER"
+    "type": "FLOAT"
   },
   {
     "mode": "NULLABLE",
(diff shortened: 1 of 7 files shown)


erika-sdf (approved):

schemas/accounts_schema.json
```diff
@@ -69,9 +69,29 @@
     "name": "last_modified_ledger",
     "type": "INTEGER"
   },
+  {
+    "mode": "NULLABLE",
+    "name": "ledger_entry_change",
```
tamirms, line 74: why do the `ledger_entry_change` fields have an integer type?

sydneynotthecity: I was taking the raw `LedgerEntryChangeType` from the [XDR](https://pkg.go.dev/github.com/stellar/go@v0.0.0-20211103162123-89aa5386377f/xdr#LedgerEntryChangeType), which was an int (enum). The ints correspond to the integer representations of the change types. Do you think it should be typed differently?

sydneynotthecity: That's fair. Since we would have to touch multiple codebases, i've created an issue for @erika-sdf and myself to tackle to make the change. 
[#215](https://github.com/stellar/hubble/issues/215)

PR #76, opened 2021-11-08
Add liquidity pools to DAGs

sydneynotthecity:
The `liquidity_pools` tables can be populated through either the use of the `bounded_core` DAG or the `bucket_list` DAG (used for initial table hydration).

### DIFF
--- a/dags/stellar_etl_airflow/build_load_task.py
+++ b/dags/stellar_etl_airflow/build_load_task.py
@@ -111,7 +111,10 @@ def upload_to_gcs(data_type, prev_task_id, batch_stats, **kwargs):
     filename = kwargs['task_instance'].xcom_pull(task_ids=prev_task_id)
     logging.info(type(filename))
     if isinstance(filename, dict):
-        filename = filename["output_file"]
+        if 'start' in filename.keys():
+            filename = f'{filename["start"]}-{filename["start"]+63}-{data_type}.txt'
+        else:
+            filename = filename["output_file"]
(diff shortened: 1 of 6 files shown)


dags/stellar_etl_airflow/build_load_task.py
```diff
@@ -111,7 +111,10 @@ def upload_to_gcs(data_type, prev_task_id, batch_stats, **kwargs):
     filename = kwargs['task_instance'].xcom_pull(task_ids=prev_task_id)
     logging.info(type(filename))
```
erika-sdf, line 112: This looks like something leftover from debugging. Should we remove it?

dags/stellar_etl_airflow/build_load_task.py
```diff
@@ -111,7 +111,10 @@ def upload_to_gcs(data_type, prev_task_id, batch_stats, **kwargs):
...
     if isinstance(filename, dict):
-        filename = filename["output_file"]
+        if 'start' in filename.keys():
+            filename = f'{filename["start"]}-{filename["start"]+63}-{data_type}.txt'
```
erika-sdf, line 115: nit: Can this entire line be pulled into a function that builds the filename?

erika-sdf (approved):
A small nit, but otherwise LGTM.

dags/stellar_etl_airflow/build_load_task.py
sydneynotthecity: Done!

PR #78, opened 2021-11-19
Update dags to use --gcs_bucket instead of load tasks

erika-sdf:
This PR considers 4 DAGS:

* `history_archives` (without CaptiveCore)
* `history_archives_with_captive_core`: `history_trades` and `history_operations` split into its own DAG
* `bounded_captive_core` for state tables 
* `bucket_list` -- @sydneynotthecity: how is this used? is this different from `bounded_captive_core`?

This change requires giving GCS permissions to `471979297599-compute@developer.gserviceaccount.com` for the GCS destination. I won't do this now so as not to disturb prod, but this will need to happen when it's rolled out.

### DIFF
--- a/airflow_variables.txt
+++ b/airflow_variables.txt
@@ -4,6 +4,7 @@
     "bq_dataset": "test_gcp_airflow_internal",
     "bq_project": "test-hubble-319619",
     "gcs_exported_data_bucket_name": "us-central1-internal-airflo-8d08728f-bucket",
+    "gcs_exported_object_prefix": "dag-exported",
     "image_name": "stellar/stellar-etl:latest",
     "image_output_path": "/etl/exported_data/",
     "image_pull_policy": "Always",
@@ -20,6 +21,7 @@
         "dimOffers": "dimOffers.txt",
         "factEvents": "factEvents.txt",
         "ledgers": "ledgers.txt",
+        "liquidity_pools": "liquidity_pools.txt",
         "offers": "offers.txt",
         "operations": "operations.txt",
         "orderbooks": "orderbook_folder",
@@ -38,13 +40,14 @@
     },
     "schema_filepath": "/home/airflow/gcs/dags/schemas/",
     "table_ids": {
-        "accounts": "acccounts",
+        "accounts": "accounts",
         "assets": "history_assets",
         "dimAccounts": "dim_accounts",
         "dimMarkets": "dim_markets",
         "dimOffers": "dim_offers",
         "factEvents": "fact_offer_events",
         "ledgers": "history_ledgers",
+        "liquidity_pools": "liquidity_pools",
         "offers": "offers",
         "operations": "history_operations",
         "trades": "history_trades",
@@ -52,6 +55,7 @@
         "trustlines": "trust_lines"
     },
     "use_kubernetes_pod_exporter": "False",
+    "use_testnet": "False",
     "volume_config": "{}",
     "volume_name": "etl-data"
 }
--- a/airflow_variables.txt
+++ b/airflow_variables.txt
@@ -54,7 +54,6 @@
         "transactions": "history_transactions",
         "trustlines": "trust_lines"
     },
-    "use_kubernetes_pod_exporter": "False",
     "use_testnet": "False",
     "volume_config": "{}",
     "volume_name": "etl-data"
--- a/dags/bounded_core_dag.py
+++ b/dags/bounded_core_dag.py
@@ -1,22 +1,17 @@
 '''
-The bounded_core_export DAG exports ledger entry changes (accounts, offers, and trustlines) within a bounded range using stellar-core. 
-This DAG should be triggered manually if it is required to export entry changes within a specified time range. For consistent updates, use
-the unbounded version.
+The state_table_export DAG exports ledger entry changes (accounts, offers, and trustlines) within a bounded range using stellar-core. 
+This DAG should be triggered manually if it is required to export entry changes within a specified time range. 
 '''
+import ast
+import datetime
 import json
-from time import time
+import logging
 
 from stellar_etl_airflow.build_export_task import build_export_task
 from stellar_etl_airflow.build_time_task import build_time_task
 from stellar_etl_airflow.default import get_default_dag_args
 from stellar_etl_airflow.build_batch_stats import build_batch_stats
 from stellar_etl_airflow.build_gcs_to_bq_task import build_gcs_to_bq_task
-
-import datetime
-import distutils
-import distutils.util
-import logging
-import time
 
 from airflow import DAG
 from airflow.models import Variable
@@ -26,15 +21,15 @@ logger = logging.getLogger('airflow.task')
 logger.setLevel(logging.INFO)
 
 dag = DAG(
-    'bounded_core_export',
+    'state_table_export',
     default_args=get_default_dag_args(),
-    start_date=datetime.datetime(2021, 10, 14),
+    start_date=datetime.datetime(2021, 11, 22, 3),#datetime.datetime(2021, 11, 29, 20),
     description='This DAG runs a bounded stellar-core instance, which allows it to export accounts, offers, liquidity pools, and trustlines to BigQuery.',
     schedule_interval='*/30 * * * *',
     user_defined_filters={'fromjson': lambda s: json.loads(s)},
 )
 
-use_testnet = bool(distutils.util.strtobool(Variable.get("use_testnet")))
+use_testnet = ast.literal_eval(Variable.get("use_testnet"))
 file_names = Variable.get('output_file_names', deserialize_json=True)
 
 date_task = build_time_task(dag)
--- a/dags/bounded_core_dag.py
+++ b/dags/bounded_core_dag.py
@@ -12,26 +12,34 @@ from stellar_etl_airflow.default import get_default_dag_args
 from stellar_etl_airflow.build_batch_stats import build_batch_stats
 from stellar_etl_airflow.build_load_task import build_load_task
 from stellar_etl_airflow.build_gcs_to_bq_task import build_gcs_to_bq_task
+
 import datetime
+import distutils
+import distutils.util
+import logging
 import time
 
 from airflow import DAG
 from airflow.models import Variable
+
+logging.basicConfig(format='%(message)s')
+logger = logging.getLogger('airflow.task')
+logger.setLevel(logging.INFO)
 
 dag = DAG(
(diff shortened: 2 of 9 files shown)


erika-sdf:
This is [currently failing](https://pf66716d511417cb1p-tp.appspot.com/admin/airflow/log?task_id=export_trades_task&dag_id=history_archive_export_with_captive_core&execution_date=2021-11-18T05%3A30%3A00%2B00%3A00&format=json) on the step `export_trades` in dev because turning on `--strict_export` causes the task to crash instead of skipping over the failed entry. 

The error message for this failure looks like:
```
from ledger 1049244, transaction 3, operation 0: Both Selling and Buying amount are 0 for operation at index 0  
```

In a [prod task](https://jf217ff1bf626397ap-tp.appspot.com/admin/airflow/log?task_id=export_trades_task&dag_id=history_archive_trades_load&execution_date=2021-11-19T22%3A00%3A00%2B00%3A00&format=json), you can see a number of similar failure messages that are skipped over since `--strict_export` is not set.

It seems like these entries are failing [at this check](https://github.com/stellar/stellar-etl/blob/12a27f07068fd588137220e423c7d931f4473d43/internal/transform/trade.go#L64) on operations where `amount` is 0.

Here are examples of a failing operation in [dev](https://horizon-testnet.stellar.org/transactions/80ff2bddb22b8b064d58fc57d0b377e78e11eaa1d1fa02694daa642704889176/operations) and [prod](https://horizon.stellar.org/operations/164765176086671361).

I think the fix here is to modify `transform/trade.go` and remove the failing check, but wanted to check in to see if it's safe/the right thing to do before doing so.

sydneynotthecity:
@erika-sdf I agree with this approach. Please modify the `transform/trade.go` to either log the message at the `Info` level (maybe even the `Debug` level.) This is expected behavior based on how `stellar-core` garbage collects invalid offers (see [Horizon note](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/processors/trades_processor.go#L251-L260) )

These messages are really noisy in the production instance and clutter the logs in my opinion.

sydneynotthecity:
`bucket_list` DAG has unique functionality compared to `bounded_captive_core`. I think we could figure out a way to collapse into one DAG and pass a flag for bucket vs bounded if we wanted.

`bucket_list` - given an end ledger, the DAG will create a `NewCheckpointChangeReader` that ingests ledgers from the Genesis Ledger until the closest checkpoint ledger to the end ledger. It uses the history buckets to build and catchup history without reading each ledger change, which means that it does not capture historical changes. Each record will be unique for its grain, with only the `last_modified_ledger` state for the record written to the table. 

`bounded_core` - needs both a start and end time to bound the range of ledgers ingested using a `LedgerChangeReader` and `LedgerTransactionReader`. Any state changes to a unique record will be recorded if it occurs during the bounded window, meaning that unique records must be a combination of the natural keys + `last_modified_ledger`.

To highlight the difference between the DAGs, if the `bucket_list` DAG was passed an end time of `2021-11-01`, and an account was created in `2018-01-01`, only the last account state would be loaded to the table and no changes between account creation and current state would be captured. Alternatively, the `bounded_core` DAG would capture every state change if you ran a batch for `2021-10-31` to `2021-11-01`, resulting in multiple records for the account if there was account activity during the given time period.

dags/bounded_core_dag.py
```diff
@@ -43,31 +50,17 @@
...
 Entries are updated, deleted, or inserted as needed.
 '''
-send_acc_to_bq_task = build_gcs_to_bq_task(dag, 'accounts', partition=False)
-send_off_to_bq_task = build_gcs_to_bq_task(dag, 'offers', partition=False)
-send_pool_to_bq_task = build_gcs_to_bq_task(dag, 'liquidity_pools', partition=False)
-send_trust_to_bq_task = build_gcs_to_bq_task(dag, 'trustlines', partition=False)
+send_acc_to_bq_task = build_gcs_to_bq_task(dag, changes_task.task_id, 'accounts', '/*-accounts.txt', partition=False)
+send_off_to_bq_task = build_gcs_to_bq_task(dag, changes_task.task_id, 'offers', '/*-offers.txt', partition=False)
+send_pool_to_bq_task = build_gcs_to_bq_task(dag, changes_task.task_id,'liquidity_pools', '/*-liquidity_pools.txt', partition=False)
+send_trust_to_bq_task = build_gcs_to_bq_task(dag, changes_task.task_id, 'trustlines', '/*-trustlines.txt', partition=False)
```
sydneynotthecity, line 61: ❓ Is the GCS file location specific enough where we do not need to worry about duplicate loading of files by using the `/*-` prefix?

dags/history_archive_with_captive_core_dag.py
```diff
@@ -0,0 +1,74 @@
...
+from stellar_etl_airflow.build_gcs_to_bq_task import build_gcs_to_bq_task
+
+from airflow import DAG
+from airflow.models import Variable
+
+
+dag = DAG(
+    'history_archive_export_with_captive_core',
+    default_args=get_default_dag_args(),
+    start_date=datetime.datetime(2021, 10, 14),
```
sydneynotthecity, line 24: Note: when we deploy to production, we'll need to put an end time on current operations/trades DAGs and new `start_date` on this DAG so we can retire those DAGs in favor of this one

sydneynotthecity (commented):
Couple questions, otherwise LGTM! I want to check out the DAG in test and the spot check the data in BQ prior to approval

dags/history_archive_with_captive_core_dag.py
erika-sdf: I've updated the dags to have appropriate `start_date`s and `end_date`s. In particular, I've renamed `bounded_core` and `history_archive` to avoid conflict, and `bucket_list_export` have an inactive `end_date`.

Will check in on the dates again before submitting to make sure they reflect the deployment start/end date.

dags/bounded_core_dag.py
erika-sdf: The destination location will be something like:
```
gs://<vars.gcs_bucket_name>/<vars.gcs_exported_object_prefix>/<run_id>/<output_path>
```

Examples from a dev run:
```
gs://us-central1-test-airflow-190ae7dc-bucket/erika-test/scheduled__2021-10-15T00:00:00+00:00/486569-486569-accounts.txt
gs://us-central1-test-airflow-190ae7dc-bucket/erika-test/scheduled__2021-10-15T00:00:00+00:00/486569-486569-offers.txt
gs://us-central1-test-airflow-190ae7dc-bucket/erika-test/scheduled__2021-10-15T00:00:00+00:00/486569-486569-trustlines.txt
```

There is a possible collision if <run_ids> are shared across dags, and the dags produce the same resource type (eg. `trustlines`, `accounts`, etc), but I believe we decided that each resource type is only produced by one dag.

airflow_variables.txt
```diff
@@ -38,20 +40,22 @@
...
         "factEvents": "fact_offer_events",
         "ledgers": "history_ledgers",
+        "liquidity_pools": "liquidity_pools",
         "offers": "offers",
         "operations": "history_operations",
         "trades": "history_trades",
         "transactions": "history_transactions",
         "trustlines": "trust_lines"
     },
     "use_kubernetes_pod_exporter": "False",
```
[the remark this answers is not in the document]
erika-sdf: Good point. I've removed it from the variables.

dags/bounded_core_dag.py
```diff
@@ -10,28 +10,35 @@
 from stellar_etl_airflow.build_time_task import build_time_task
 from stellar_etl_airflow.default import get_default_dag_args
 from stellar_etl_airflow.build_batch_stats import build_batch_stats
-from stellar_etl_airflow.build_load_task import build_load_task
 from stellar_etl_airflow.build_gcs_to_bq_task import build_gcs_to_bq_task
+
 import datetime
+import distutils
+import distutils.util
```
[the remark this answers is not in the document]
erika-sdf: Good catch. I've replaced the references to `disutils`.

erika-sdf:
> `bucket_list` DAG has unique functionality compared to `bounded_captive_core`. I think we could figure out a way to collapse into one DAG and pass a flag for bucket vs bounded if we wanted.
> 
> `bucket_list` - given an end ledger, the DAG will create a `NewCheckpointChangeReader` that ingests ledgers from the Genesis Ledger until the closest checkpoint ledger to the end ledger. It uses the history buckets to build and catchup history without reading each ledger change, which means that it does not capture historical changes. Each record will be unique for its grain, with only the `last_modified_ledger` state for the record written to the table.
> 
> `bounded_core` - needs both a start and end time to bound the range of ledgers ingested using a `LedgerChangeReader` and `LedgerTransactionReader`. Any state changes to a unique record will be recorded if it occurs during the bounded window, meaning that unique records must be a combination of the natural keys + `last_modified_ledger`.
> 
> To highlight the difference between the DAGs, if the `bucket_list` DAG was passed an end time of `2021-11-01`, and an account was created in `2018-01-01`, only the last account state would be loaded to the table and no changes between account creation and current state would be captured. Alternatively, the `bounded_core` DAG would capture every state change if you ran a batch for `2021-10-31` to `2021-11-01`, resulting in multiple records for the account if there was account activity during the given time period.
> 
> The `bucket_list` functionality is only useful for creating a starting point/incomplete historical load for the state tables and should not be used for anything beyond ad hoc historical loads.

Thanks for the explanation. Sounds like we'll keep around all 4 of the dags then, but I won't turn `export_bucket` on.

> @erika-sdf I agree with this approach. Please modify the `transform/trade.go` to either log the message at the `Info` level (maybe even the `Debug` level.) This is expected behavior based on how `stellar-core` garbage collects invalid offers (see [Horizon note](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/processors/trades_processor.go#L251-L260) )
> 
> These messages are really noisy in the production instance and clutter the logs in my opinion.

Cool. I've send you a [PR](https://github.com/stellar/stellar-etl/pull/147) with this change.

dags/history_archive_dag.py
```diff
@@ -18,14 +18,16 @@
 
 
 dag = DAG(
-    'history_archive_export',
+    'history_archive_without_captive_core',
```
sydneynotthecity, line 21: Let's plan to upload this DAG to the same airflow instance and then we can delete the old one.

erika-sdf: Alright, I'll aim to move everything into the `bounded-captive-core` airflow instance.

PR #79, opened 2021-11-30
Increase max bad records allowed in BQ

sydneynotthecity:
There is a [bug](https://github.com/stellar/hubble/issues/230) in the current `history_operations` table that will need to be fixed. Currently, any `*_muxed_id` field is defined as an `INTEGER`, but the id can be longer than the max integer. 

As a workaround, we are increasing the `MAX_BAD_RECORDS` to 10 records per batch. It is better to ignore the bad record and insert the remaining records in the batch than to reject the entire batch of records.


erika-sdf (approved):
LGTM

PR #81, opened 2021-12-15
Add claimable balances to state_table_dag

erika-sdf:


dags/claimable_balances_history_dag.py
```diff
@@ -0,0 +1,59 @@
...
+
+logging.basicConfig(format='%(message)s')
+logger = logging.getLogger('airflow.task')
+logger.setLevel(logging.INFO)
+
+dag = DAG(
+    'claimable_balances_history',
+    default_args=get_default_dag_args(),
+    start_date=datetime.datetime(2021, 10, 1),
+    end_date=datetime.datetime(2021, 11, 29, 22, 29),
```
sydneynotthecity, line 27: Will this be changed to the deployment date? I imagine we would backfill through whatever datetime we enabled `claimable_balances` in the state table

dags/claimable_balances_history_dag.py
```diff
@@ -0,0 +1,59 @@
...
+    description='This DAG runs a bounded stellar-core instance, which allows it to export accounts, offers, liquidity pools, and trustlines to BigQuery.',
+    schedule_interval='*/30 * * * *',
```
sydneynotthecity, line 29: For backfilling we can increase this to `'0 */3 * * * '` and execute every 3 hours instead

sydneynotthecity (commented):
We are missing two more file changes: 
1. `bucket_list_dag` should be modified to include `claimable_balances`
2. We should have an ad hoc dag, that is a copy of `bucket_list_dag` with only `claimable balances` that will give a snapshot of the state for `2021-10-01` prior to backfilling all state changes

dags/claimable_balances_history_dag.py
erika-sdf: done

dags/claimable_balances_history_dag.py
erika-sdf: The end date is currently set to the start date of the `state_table_dag`, but you're right that it should be changed to the deployment date just before it's actually deployed. I'll wait on the deployment before submitting.

erika-sdf:
> We are missing two more file changes:
> 
> 1. `bucket_list_dag` should be modified to include `claimable_balances`

I've updated the `bucket_list_dag` to include `claimable_balances`.

> 2. We should have an ad hoc dag, that is a copy of `bucket_list_dag` with only `claimable balances` that will give a snapshot of the state for `2021-10-01` prior to backfilling all state changes
 
I've added a `claimable_balances_bucket_list_dag` that should run once, for `2021-10-01`

sydneynotthecity (approved):
LGTM

PR #82, opened 2022-01-04
Add field to `claimants` in claimable balances

sydneynotthecity:
Updating the Hubble table schemas to add `abs_before_epoch` optional field to the `claimants` object. 

This field is the integer representation of the ISO date found in `abs_before` field. New field aligns with `rel_before` formatting.

Claimants are found on both `history_operations` and `claimable_balances` tables, so both schemas must be updated.

Update fixes breaking [schema change](https://github.com/stellar/go/pull/4148/files) from Horizon release


erika-sdf (approved):
LGTM, nice that this was addressable with only a schema change.

PR #83, opened 2022-01-05
Split out dev and prod env vars and tag image version

erika-sdf:
Tag images with a unique tag when built, set hubble to reference a pinned image.

### DIFF
--- a/airflow_variables_dev.txt
+++ b/airflow_variables_dev.txt
@@ -1,9 +1,9 @@
 {
     "affinity": "{}",
     "api_key_path": "/home/airflow/gcs/data/apiKey.json",
-    "bq_dataset": "crypto_stellar_internal_2",
-    "bq_project": "hubble-261722",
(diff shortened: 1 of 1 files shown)


airflow_variables_dev.txt
```diff
@@ -0,0 +1,62 @@
...
+    },
+    "schema_filepath": "/home/airflow/gcs/dags/schemas/",
+    "table_ids": {
+        "accounts": "accounts",
+        "assets": "history_assets",
+        "claimable_balances": "claimable_balances",
+        "dimAccounts": "dim_accounts",
+        "dimMarkets": "dim_markets",
+        "dimOffers": "dim_offers",
+        "factEvents": "fact_offer_events",
```
sydneynotthecity, line 50: Should we also clean these up since we're not using the orderbook code right now?

sydneynotthecity (approved):
One nit, looks good!

airflow_variables_dev.txt
erika-sdf: Removed the orderbook table names

PR #86, opened 2022-01-05
Remove unused dags and tasks

erika-sdf:


dags/stellar_etl_airflow/build_load_task.py
```diff
@@ -1,196 +0,0 @@
-'''
```
sydneynotthecity, line 1: Can `stellar-etl` still write files only locally instead upload to GCS (maybe AWS in the future)? 

I'm assuming if we needed to upload somewhere beyond GCS we would build that functionality into the docker image itself and not the DAG?

sydneynotthecity (approved):
One question, overall LGTM

dags/stellar_etl_airflow/build_load_task.py
erika-sdf: Yes, not specifying the GCS flag will write it to disk. (Changes would be needed for supporting AWS). I do think it's generally easier to do the desired upload behavior in the docker image.

PR #88, opened 2022-02-11
Switch build_export_task from DockerOperator to KubernetesPodOperator

erika-sdf:


dags/stellar_etl_airflow/build_export_task.py
```diff
@@ -135,5 +108,23 @@ def build_export_task(dag, cmd_type, command, filename, use_gcs=False, use_testn
...
+        arguments=[f'''{etl_cmd_string} && echo "{{\\"output\\": \\"{output_file}\\"}}" >> /airflow/xcom/return.json'''],
+        dag=dag,
+        do_xcom_push=True,
+        is_delete_operator_pod=True,
+        in_cluster=in_cluster,
+        config_file=config_file_location,
+        affinity=Variable.get('affinity', deserialize_json=True),
+        auto_remove=True,
+        tty=True,
+        force_pull=force_pull,
```
sydneynotthecity, line 129: I think we also have the option to use `image_pull_policy` instead of `force_pull` but that might be new versions only

sydneynotthecity (approved):
LGTM!

dags/stellar_etl_airflow/build_export_task.py
erika-sdf: Good point. I swapped it in

PR #89, opened 2022-02-22
Update Kube Settings

sydneynotthecity:
The `KubernetesPodOperator` will now pull the resources for kube pods from the `airflow.txt` resources parameter. Pods should be bigger to accommodate the CaptiveCore size.


erika-sdf (approved):
Seems like this could help.

PR #91, opened 2022-03-09
Add affinity for kube operators

sydneynotthecity:
The airflow scheduler was experiencing resource starvation because CaptiveCore tasks were executing on the same pods that the scheduler was. The PR adds node pools so that process intensive tasks execute separate from Airflow infrastructure.


erika-sdf (approved):
LGTM

PR #92, opened 2022-03-10
Upgrade Airflow 1.10 -> 2.2.3

erika-sdf:


README.md
```diff
@@ -57,6 +57,11 @@ gcloud composer environments create <environment_name> --location=<project_locat
 gcloud composer environments update <environment_name> \
 --location=<project_location> --update-pypi-package=docker==3.7.3
 ```
+
+> **_For AIRFLOW 2.x:_** Be wary of choosing "autopilot" for environment resource management. The ephemeral storage provided by autopilot-ed containers is capped at 10GB, which may not be enough for hefty tasks (such as `state_table_dag`'s `export_task`).
```
sydneynotthecity, line 61: Can we add a note that CaptiveCore needs more than 10GB to execute?

README.md
```diff
@@ -57,6 +57,11 @@ gcloud composer environments create <environment_name> --location=<project_locat
...
+You can add a second node pool to your Composer 1 environment, and configure it  be managed by autopilot if desired.
```
sydneynotthecity, line 62: > configure it `to` be

airflow.cfg
```diff
@@ -0,0 +1,81 @@
...
+base_log_folder = /home/airflow/gcs/logs
+remote_logging = True
+remote_log_conn_id = google_cloud_default
+remote_base_log_folder = gs://us-central1-hubble-1pt5-dev-7db0e004-bucket/logs
+
+[webserver]
+web_server_host = 0.0.0.0
+web_server_port = 8080
+secret_key = temporary_key
+workers = 2
```
sydneynotthecity, line 31: Are these the `airflow-worker` pods? or different pods?

airflow_variables_dev.txt
```diff
@@ -34,14 +67,25 @@
...
-        "request_memory": "5Gi",
-        "limit_memory": "15Gi",
-        "request_cpu": "1.2",
-        "limit_ephemeral_storage": "12Gi"
+        "default": {
+            "request_cpu": "1.2",
+            "request_memory": "5Gi",
+            "request_ephemeral_storage": "10Mi"
+        },
+        "cc": {
```
sydneynotthecity, line 75: Whats the difference between `cc` and `state`? If both presumably use a `CaptiveCore` instance, why do the resources have to be defined differently between the two?

dags/stellar_etl_airflow/build_batch_stats.py
```diff
@@ -9,7 +9,7 @@ def build_batch_stats(dag, table):
     start_ledger = '{{ ti.xcom_pull(task_ids="get_ledger_range_from_times")["start"] }}'
     end_ledger = '{{ ti.xcom_pull(task_ids="get_ledger_range_from_times")["end"] - 1 }}'
     batch_id = '{{ run_id }}'
-    batch_run_date = '{{ prev_execution_date.to_datetime_string() }}'
+    batch_run_date = '{{ logical_date.to_datetime_string() }}'
```
sydneynotthecity, line 12: Nice catch 💯 

Does `logical_date` work by itself or do you need to reference the dag_run: `dag_run.logical_date`?

dags/stellar_etl_airflow/build_export_task.py
```diff
@@ -107,26 +107,28 @@ def build_export_task(dag, cmd_type, command, filename, use_gcs=False, use_testn
...
         dag=dag,
         do_xcom_push=True,
         is_delete_operator_pod=True,
         startup_timeout_seconds=720,
-        resources=ast.literal_eval(Variable.get('resources')),
+        resources=resources,
         in_cluster=in_cluster,
         config_file=config_file_location,
-        affinity=Variable.get('affinity', deserialize_json=True),
-        auto_remove=True,
```
sydneynotthecity, line 129: Do we need to add `is_delete_operator_pod` and set to `True`? The `auto_remove` parm would delete the pod after execution. As long as we are preserving the same functionality I'm good with it.

dags/stellar_etl_airflow/kubernetes_pod_operator.py
```diff
@@ -1,410 +0,0 @@
-"""
```
sydneynotthecity, line 1: Thanks for cleaning these up!

sydneynotthecity (approved):
Couple questions. Overall looks good! Good work on the upgrade 😄

README.md
erika-sdf: Added the detail that the requirements are insufficient for captive core jobs.

airflow_variables_dev.txt
erika-sdf: The state job historically uses slightly more ephemeral storage. 

Here's a [link to a graph](https://console.cloud.google.com/monitoring/metrics-explorer?project=test-hubble-319619&pageState=%7B%22xyChart%22:%7B%22dataSets%22:%5B%7B%22timeSeriesFilter%22:%7B%22filter%22:%22metric.type%3D%5C%22kubernetes.io%2Fcontainer%2Fephemeral_storage%2Fused_bytes%5C%22%20resource.type%3D%5C%22k8s_container%5C%22%20resource.label.%5C%22container_name%5C%22%3D%5C%22base%5C%22%20resource.label.%5C%22cluster_name%5C%22%3D%5C%22us-central1-test-airflow-190ae7dc-gke%5C%22%22,%22minAlignmentPeriod%22:%2260s%22,%22aggregations%22:%5B%7B%22perSeriesAligner%22:%22ALIGN_MEAN%22,%22crossSeriesReducer%22:%22REDUCE_NONE%22,%22alignmentPeriod%22:%2260s%22,%22groupByFields%22:%5B%5D%7D,%7B%22perSeriesAligner%22:%22ALIGN_NONE%22,%22crossSeriesReducer%22:%22REDUCE_NONE%22,%22alignmentPeriod%22:%2260s%22,%22groupByFields%22:%5B%5D%7D%5D%7D,%22targetAxis%22:%22Y1%22,%22plotType%22:%22LINE%22%7D%5D,%22options%22:%7B%22mode%22:%22COLOR%22%7D,%22constantLines%22:%5B%5D,%22timeshiftDuration%22:%220s%22,%22y1Axis%22:%7B%22label%22:%22y1Axis%22,%22scale%22:%22LINEAR%22%7D%7D,%22isAutoRefresh%22:true,%22timeSelection%22:%7B%22timeRange%22:%226w%22%7D,%22xZoomDomain%22:%7B%22start%22:%222022-03-07T23:51:28.644Z%22,%22end%22:%222022-03-08T23:14:51.667Z%22%7D%7D) that shows ephemeral storage usage by task name (in dev, but I think we can expect similar usage pattern). You can see that the state export tasks sometimes spike up to 12GB, while the others don't quite surpass 10GB.

I have not be able to pinpoint why peak usage is about 2GB higher, but would love insight if you have any.

Resource usage will probably change as we continue to scale, but for now, the current numbers in the config seem sufficient to keep tasks from being ejected for the time being.

We can merge this into a single definition with the higher requirements if you'd like to simplify the config.

dags/stellar_etl_airflow/build_batch_stats.py
erika-sdf: This seems to be working on it's own

dags/stellar_etl_airflow/build_export_task.py
erika-sdf: I was under the impression this was being deprecated, but cannot find evidence of that now. I will add `is_delete_operator_pod`.

airflow_variables_dev.txt
sydneynotthecity: it's fine as is and we can continue to monitor. I wonder if the 2GB difference is due to the fact that we're extracting all state tables vs just the `history_operations` or `history_trades`. It's possible that the extracted state table data accounts for that increase?? Keeping the resource requirements separate is fine, was more curious why `state` seems to use more.

PR #98, opened 2022-04-04
Enable public dataset writing

sydneynotthecity:
In an effort to retire the Hubble 1.0 pipeline, we need the ability to write the transformed `stellar-etl` data to public datasets. Namely, `crypto-stellar:crypto_stellar`. The easiest way to do this is by creating an additional `GCSToBigquery` airflow task that writes the dataset to a publicly specified location. 

For the public dataset, I'd like to introduce clustering to our tables to create efficiencies in complex queries. By clustering fields based on common join keys, we can add granularity in the table organization beyond partitioning. In a future PR, we should work on making the Hubble 2.0 tables backwards compatible so that they are also clustered on the same fields.

This PR covers those changes:

* Adding a `public_project` and `public_dataset` parameter 
* Creating new Airflow tasks that will write the data to the public datasets
* Modifying the DAG definition so that two tasks run _in parallel_ after `delete_<name>_task`
* Updating the `create_tables` DDL script for optimized table format
* Allowing tables to contain `cluster_fields` which are defined in the `airflow_variables.txt` parameter file.

### DIFF
--- a/airflow_variables.txt
+++ b/airflow_variables.txt
@@ -35,6 +35,19 @@
     "api_key_path": "/home/airflow/gcs/data/apiKey.json",
     "bq_dataset": "crypto_stellar_internal_2",
     "bq_project": "hubble-261722",
+    "cluster_fields": { 
+        "accounts": ["account_id","last_modified_ledger"], 
+        "account_signers": ["account_id","last_modified_ledger"], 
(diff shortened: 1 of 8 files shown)


erika-sdf (approved):
LGTM

PR #99, opened 2022-04-26
Add enriched hist ops

sydneynotthecity:
PR automates the update of `enriched_history_operations` table and improves the pipeline to incrementally load the table based on batch id. The pipeline is contained within the `history_archive_with_captive_core_dag` and waits on the `history_operations` table to finish. The pipeline adds the following steps: 

1. Wait on the other DAG, `history_archive_without_captive_core_dag` to finish successfully prior to starting
2. Run the query that normalizes ledgers, ops, txs to the `enriched_history_operations` table. The query is parameterized by project, dataset, and batch id so that only the latest batch is executed and appended.
3. Insert the records using `WRITE_APPEND` write disposition to both public and internal datasets.

Other improvements: 

- Change the `build_delete_data_task.py` to delete data from **both** the public and internal datasets. Task now expects the user to pass the `project` and `dataset` ids as parameters.
- Add the view definition of `v_accounts_current`
- Update all DAGs to delete the public dataset partition prior to rerunning so that duplicate partitions are not inserted into the tables.


erika-sdf (approved):
Awesome, nice to see these improvements. Does this resolve the issue of missing trade entries?

sydneynotthecity:
It does not resolve the missing trade entries. I'm working on that next. This just makes sure we have coverage for derived Hubble 1.0 table with accompanying data controls (like deleting duplicate partitions)

PR #100, opened 2022-05-06
Add enriched meaningful hist ops table

sydneynotthecity:
Adds derivation code and data population for `enriched_meaningful_history_operations` table. The table filters down the `enriched_history_operations` down to operations that interacted with relevant assets.

The pipeline follows the framework of the `enriched_history_operations` table:
1. Check if the current `batch_id` partition exists in the table. If it does, delete the old one
2. Execute a query parameterized by `batch_id` that will insert fresh data from `enriched_history_operations` table into the final table.

### DIFF
--- a/dags/history_archive_with_captive_core_dag.py
+++ b/dags/history_archive_with_captive_core_dag.py
@@ -81,6 +81,7 @@ delete_old_trade_task = build_delete_data_task(dag, internal_project, internal_d
 delete_old_trade_pub_task = build_delete_data_task(dag, public_project, public_dataset, table_names['trades'])
 delete_enrich_op_task = build_delete_data_task(dag, internal_project, internal_dataset, 'enriched_history_operations')
 delete_enrich_op_pub_task = build_delete_data_task(dag, public_project, public_dataset, 'enriched_history_operations')
+delete_enrich_ma_op_task = build_delete_data_task(dag, internal_project, internal_dataset, 'enriched_meaningful_history_operations')
 
 '''
 The send tasks receive the location of the file in Google Cloud storage through Airflow's XCOM system.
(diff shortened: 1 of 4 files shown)


sydneynotthecity:
PR is not time sensitive but several [Metabase dashboards](https://docs.google.com/spreadsheets/d/1Bl9ewIcHqn2KbDm0U85eTNHLDPLHsWLnsO18RQTu9mQ/edit#gid=0) depend on table creation in order to be eligible for migration.

erika-sdf (approved):

dags/queries/enriched_meaningful_history_operations.sql
```diff
@@ -0,0 +1,20 @@
...
+*/
+SELECT
+  *
+FROM `{project_id}.{dataset_id}.enriched_history_operations` eho
+INNER JOIN `{project_id}.{dataset_id}.meaningful_assets` ma ON
+  (eho.asset_code = ma.code AND eho.asset_issuer = ma.issuer) OR
+  (eho.source_asset_code = ma.code AND eho.source_asset_issuer = ma.issuer) OR
+  (eho.selling_asset_code = ma.code AND eho.selling_asset_issuer = ma.issuer) OR
+  (eho.buying_asset_code = ma.code AND eho.buying_asset_issuer = ma.issuer)
+WHERE (eho.successful=true OR eho.successful IS NOT NULL)
```
Shaptic, line 18: Is this conditional redundant? I would imagine that not being null includes being true :thinking:

dags/history_archive_with_captive_core_dag.py
```diff
@@ -99,12 +100,14 @@
...
 Must wait on history_archive_without_captive_core_dag to finish before beginning the job.
+The internal dataset also creates a filtered table, `enriched_meaningful_history_operations` which filters down to only relevant asset ops.
 '''
 wait_on_dag = build_cross_deps(dag, "wait_on_ledgers_txs", "history_archive_without_captive_core")
 insert_enriched_hist_task = build_bq_insert_job(dag, internal_project, internal_dataset, "enriched_history_operations", partition=True)
 insert_enriched_hist_pub_task = build_bq_insert_job(dag, public_project, public_dataset, "enriched_history_operations", partition=True)
+insert_enriched_ma_hist_task = build_bq_insert_job(dag, internal_project, internal_dataset, "enriched_meaningful_history_operations", partition=True)
 
-time_task >> write_op_stats >> op_export_task >> delete_old_op_task >> send_ops_to_bq_task >> wait_on_dag >> insert_enriched_hist_task
+time_task >> write_op_stats >> op_export_task >> delete_old_op_task >> send_ops_to_bq_task >> wait_on_dag >> insert_enriched_hist_task >> delete_enrich_ma_op_task >> insert_enriched_ma_hist_task
```
Shaptic, line 110: I have no idea what I'm looking at so this might be the weirdest Python code I've ever read. Is there some sort of freaky overloaded `__lsh__` on these variables?

dags/stellar_etl_airflow/build_time_task.py
```diff
@@ -23,7 +23,6 @@ def build_time_task(dag, use_testnet=False, use_next_exec_time=True, resource_cf
     end_time = '{{ subtract_data_interval(dag, data_interval_end).isoformat() }}' if use_next_exec_time else '{{ ts }}'
     command = ["stellar-etl"]
     args = [ "get_ledger_range_from_times", "-s", start_time, "-o", "/airflow/xcom/return.json", '-e', end_time]
-    logging.info(f'Constructing command with args: {args}')
```
Shaptic, line 26: was it annoying or something? :laughing: maybe `logging.debug`, instead?

Shaptic (approved):
I have no idea what I'm reviewing but this seems good :+1:

dags/queries/enriched_meaningful_history_operations.sql
sydneynotthecity: I thought the same thing 😖 this logic was original to Hubble 1.0 and I didn't know if there was a reason for its existence. I'll update if it's redundant (I believe it is)

dags/history_archive_with_captive_core_dag.py
Shaptic: whack; thank you for explaining!

dags/stellar_etl_airflow/build_time_task.py
Shaptic: oh makes sense, thanks for clarifying! that is obnoxious; is this meant to be run locally at all, then?

dags/queries/enriched_meaningful_history_operations.sql
sydneynotthecity: Update: deleting the `WHERE` condition for successful transaction types entirely as this will always pull 100% of transactions regardless of whether they succeeded. 

(I am also of the opinion that we would want both successful and failed transactions in this table)

dags/stellar_etl_airflow/build_time_task.py
sydneynotthecity: Only for debugging purposes when it's difficult to understand the root cause of a failure. Otherwise most testing is performed in our test Airflow instance.

PR #102, opened 2022-06-10
Improve maintainability of stellar-etl-airflow

sydneynotthecity:
Improvements to Airflow DAGs so that they are easier to maintain and monitor. 
Improvements are twofold: 

1. Incorporate an `execution_timeout` to all DAG tasks
2. Create ad hoc GCS file deletion script to delete large amounts of GCS files based on pattern matching

DAG tasks can spawn hung processes every 3-5 days that never timeout. These tasks will never complete, but require manual intervention to fail the task and restart it. The `execution_timeout` parameter sets an upper limit for a DAG task that will automatically timeout and fail the task if it exceeds the limit. This should reduce the number of manual interventions in Airflow.

In the event that a large amount of data needs to be reingested (eg, change to logic and need to backfill history), the `delete_gcs_files.sh` can be used to clear out the old GCS transformed files so that our transformed file storage can remain clean.

### DIFF
--- a/dags/stellar_etl_airflow/build_bq_insert_job_task.py
+++ b/dags/stellar_etl_airflow/build_bq_insert_job_task.py
@@ -1,4 +1,5 @@
 import os
+from datetime import timedelta
 # import config
 from airflow.models import Variable
 from airflow.providers.google.cloud.operators.bigquery import BigQueryInsertJobOperator
@@ -43,6 +44,7 @@ def build_bq_insert_job(dag, project, dataset, table, partition):
 
     return BigQueryInsertJobOperator(
         task_id=f"insert_records_{table}_{dataset_type}",
+        execution_timeout=timedelta(seconds=180),
         configuration={
             "query": {
                 "query": query,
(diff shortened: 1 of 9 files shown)


dags/stellar_etl_airflow/build_bq_insert_job_task.py
```diff
@@ -1,4 +1,5 @@
 import os
+from datetime import timedelta
 # import config
```
erika-sdf, line 3: Can you also remove this commented line?

dags/stellar_etl_airflow/build_bq_insert_job_task.py
```diff
@@ -43,6 +44,7 @@ def build_bq_insert_job(dag, project, dataset, table, partition):
 
     return BigQueryInsertJobOperator(
         task_id=f"insert_records_{table}_{dataset_type}",
+        execution_timeout=timedelta(seconds=180),
```
erika-sdf, line 47: Not blocking, but it would be nice to make these timeouts configurable through the `variables*.txt` files.

erika-sdf (approved):
A couple of comments, but nothing blocking.

dags/stellar_etl_airflow/build_bq_insert_job_task.py
sydneynotthecity: I like that idea, I'll work on it

PR #109, opened 2022-10-18
Use parameter `resources` instead of `container_resources`

lucaszanotelli:
When using the parameter `container_resources` in `KubernetesPodOperator` it generates an invalid .json configuration file.

```
        "resources": {
          "limits": {   <======== that's wrong
            "requests": {
              "cpu": "6",
              "memory": "20Gi"
            }
          }
```

Using parameter `resources` fixes the issue.


sreuland:
@lucaszanotelli , can you by chance show what the corrected json format output is after change? Is the json supposed to follow the same k8s Container definition as yaml or slightly different? Just checking [k8s container spec for resource limits](https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/#specify-a-cpu-request-and-a-cpu-limit) which shows usage of `limits`

lucaszanotelli:
yeah, just removing the `limits` line and adjusting the indentation would generate a correct config file.

dags/stellar_etl_airflow/build_time_task.py
```diff
@@ -49,6 +49,6 @@ def build_time_task(dag, use_testnet=False, use_next_exec_time=True, resource_cf
          in_cluster=in_cluster,
          config_file=config_file_location,
          affinity=affinity,
-         container_resources=k8s.V1ResourceRequirements(**resources),
```
sreuland, line 52: if this resolves the output issue, than all good, I couldn't tell from brief look at [airflow docs](https://airflow.apache.org/docs/apache-airflow-providers-cncf-kubernetes/stable/_api/airflow/providers/cncf/kubernetes/operators/kubernetes_pod/index.html?highlight=container_resources) if `container_resources` and `resources` are basically configuring the same aspect, was wondering if the `resources` itself may possibly be an issue where it is assembled with the wrong structure and it's just being passed into operator here.

lucaszanotelli: i've done another commit that adjusts it to the new [`V1ResourceRequirements ` structure](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1ResourceRequirements.md).

a unit test would be a nice think to implement.

sreuland (approved):
thanks for additional refinements. am not able to determine if output is fixed from code alone without a unit test demonstrating same aspect, but if you are able to confirm from live tests, etc, than all good, thanks!

PR #111, opened 2022-10-20
Deduplicate history_assets

lucaszanotelli:
This adds `dedup_*` step in `history_archive_without_captive_core_dag` DAG to load only new records in `history_assets` table.

Closes [GH#216](https://github.com/stellar/hubble/issues/216)

This also updates the some values in `airflow_variables.txt` files to mirror the actual pipeline values.


dags/queries/history_assets.sql
```diff
@@ -0,0 +1,20 @@
+/*
+This query deduplicates the table history_assets.
+*/
+WITH raw as (
+  SELECT
+    *
+    , row_number() OVER(PARTITION BY asset_type, asset_issuer, asset_code ORDER BY batch_run_date ASC) as dedup_oldest_asset
+  FROM `{project_id}.{dataset_id}.history_assets_staging`
```
sydneynotthecity, line 8: Does the `history_assets_staging` contain all raw history asset records (like the old table)? Or does this table either truncated history/subset of the data? I could see this query getting really expensive if there is no purge process. Thoughts?

dags/stellar_etl_airflow/build_bq_insert_job_task.py
```diff
@@ -29,32 +29,41 @@ def build_bq_insert_job(dag, project, dataset, table, partition):
...
+            'writeDisposition': 'WRITE_APPEND'
+        }
+    }
     if partition:
-        partition_fields = Variable.get("partition_fields", deserialize_json=True)
-    else:
-        partition_fields = None
+        partition_fields = Variable.get('partition_fields', deserialize_json=True)
+        configuration['query']['time_partitioning'] = partition_fields[table]
+    if table == 'history_assets':
```
sydneynotthecity, line 56: Is `history_assets` the only table that uses clustering fields in the `build_bq_insert_job_task`?

sydneynotthecity (commented):
Couple questions and then I'll approve! I'm mainly wondering about the dedup query cost and if there is any way to keep that cost small. I know the `history_assets` table is not huge, but I'm curious if we think it's going to grow large enough where this would be a problem in the future.

dags/stellar_etl_airflow/build_bq_insert_job_task.py
lucaszanotelli: As we speak `history_assets` is the only table that uses clustering. But that's no reason to not generalize this. I'll update the code.

dags/queries/history_assets.sql
lucaszanotelli: It contains all raw history. We'll change this query logic so that only stores the latest extracted data in it.

lucaszanotelli:
About costs it'll add around 4 USD per month right now and we think it's not going to grow large enough to become an issue.

dags/stellar_etl_airflow/build_bq_insert_job_task.py
sydneynotthecity: thanks!

sydneynotthecity:
thanks for researching and confirming for me, @lucaszanotelli! code looks good, i'm approving

sydneynotthecity (approved):

PerriLucas:
Since there was no need to keep the base table, we'll be recreating the current table by deduplicating it entirely, then run a different query in order to only load assets not present in said deduplicated table. The batched assets will refresh each batch, and the table for comparision will always be deduplicated, so the costs should be as low as possible.

PR #114, opened 2023-01-23
Add history effects to the current pipeline

lucaszanotelli:
The backfill for the `history_effects` table is done.

This PR includes this table to the current pipeline.

Closes [#246](https://github.com/stellar/hubble/issues/246)


dags/history_archive_with_captive_core_dag.py
```diff
@@ -70,6 +71,7 @@
 '''
 op_export_task = build_export_task(dag, 'archive', 'export_operations', file_names['operations'], use_testnet=use_testnet, use_gcs=True, resource_cfg='cc')
 trade_export_task = build_export_task(dag, 'archive', 'export_trades', file_names['trades'], use_testnet=use_testnet, use_gcs=True, resource_cfg='cc')
+effects_export_task = build_export_task(dag, 'archive', 'export_effects', 'effects.txt', use_testnet=use_testnet, use_gcs=True, resource_cfg='backfill')
```
sydneynotthecity, line 74: Is this `resource_cfg` still `backfill`? Or should this variable be updated?

schemas/history_effects_schema.json
```diff
@@ -0,0 +1,475 @@
...
+      {
+        "mode": "NULLABLE",
+        "name": "starting_balance",
+        "type": "STRING"
+      },
+      {
+        "mode": "NULLABLE",
+		    "name": "seller",
+		    "type": "STRING"
+      },
```
sydneynotthecity, line 347: Formatting is slightly messed up. Can we get the `name` and `type` aligned?

schemas/history_effects_schema.json
```diff
@@ -0,0 +1,475 @@
...
+            "fields": [
+              {
+                "mode": "NULLABLE",
+                "name": "asset",
+                "type": "STRING"
+              },
+              {
+                "mode": "NULLABLE",
+                "name": "amount",
+                "type": "STRING"
```
sydneynotthecity, line 66: What is the reason behind all of the `amount` fields being of type `STRING`?

schemas/history_effects_schema.json
```diff
@@ -0,0 +1,475 @@
...
+          },
+          {
+            "mode": "NULLABLE",
+            "name": "rel_before",
+            "type": "STRING"
+          }
+        ],
+        "mode": "NULLABLE",
+        "name": "predicate",
+        "type": "RECORD"
```
sydneynotthecity, line 443: I think with nested `predicate`s we could need a `predicate` defined more like the claimable balances [predicate](https://github.com/stellar/stellar-etl-airflow/blob/master/schemas/claimable_balances_schema.json#L14-L390), but i'm not super familiar with `history_effects` so this _might_ be enough.

sydneynotthecity (approved):
Couple questions regarding the `schema.json` of the `history_effects` table. The DAG definition looks good!

dags/history_archive_with_captive_core_dag.py
lucaszanotelli: Well noted! I'll update to `cc`.

schemas/history_effects_schema.json
lucaszanotelli: It was a distraction, the type should be `INTEGER`. Now the table was created with the schema that way… To change the column type will require some juggling with BigQuery but it's nothing out of the ordinary. Would you like me to make it so that the type of the amount column `amount` is `INTEGER`?

schemas/history_effects_schema.json
sydneynotthecity: I filed a bug and if there are any failed records in the future, we can address it at that point. 

I _think_ we would be able to create a new table and use a SQL statement to recreate that field correctly and insert all of the historic records into the new table. I'm not concerned about addressing this right now though because users rarely nest a `predicate` in that manner

sydneynotthecity (approved):

schemas/history_effects_schema.json
sydneynotthecity: Yes, if we can fix the following fields in `details`, they should all be either `NUMERIC` or `INTEGER` datatypes:
* `liquidity_pool.total_shares` change to `NUMERIC`
* `liquidity_pool.total_trustlines` change to `INTEGER`
* `liquidity_pool.reserves.amount` change to `NUMERIC`
* `reserves_received.amount` change to `NUMERIC`
* `reserves_deposited.amount` change to `NUMERIC`
* `reserves_revoked.amount` change to `NUMERIC`
* `bought.amount` change to `NUMERIC`
* `sold.amount` change to `NUMERIC`
* `shares_revoked` change to `NUMERIC`
* `shares_received` change to `NUMERIC`
* `shares_redeemed` change to `NUMERIC`
* `limit` change to `NUMERIC`
* `amount` change to `NUMERIC`
* `starting_balance` change to `NUMERIC`
* `sold_amount` change to `NUMERIC`
* `bought_amount` change to `NUMERIC`

sydneynotthecity: @lucaszanotelli do you want me to create a new issue for this work?

lucaszanotelli: Yes, that would be great

sydneynotthecity:
This can be merged, however I would like the `amount` types updated from `STRING` to `NUMERIC` or `INTEGER` (see comment) in the `details` column.

Also @lucaszanotelli can we partition and cluster this table in our internal dataset? I noticed that it is not partitioned or clustered and I would like it to match the public table.

lucaszanotelli:
Ok, I'll update the files to make it partitioned and clustered.

PR #119, opened 2023-03-10
Add messages to Sentry

lucaszanotelli:
This PR adds a couple bug tracking features to Airflow DAGs with the use of Sentry SDK.

The following messages (with level as `error`) are sent to `anomaly-detection` project when:

1. any task reaches it's max retry value;
2. any export task outputs a [`failed_transforms`](https://github.com/stellar/stellar-etl/blob/2426e9310dcb024452eacdf3d701dd1b83d2c6f0/cmd/export_ledgers.go#L91) value over the max value (set as `0` for now).

This also updates the `sentry_dsn` variable to report errors to `anomaly-detection` project.

Closes issues [#194](https://github.com/stellar/hubble/issues/194) and [#300](https://github.com/stellar/hubble/issues/300)


airflow_variables_dev.json
```diff
@@ -169,7 +169,7 @@
     }
   },
   "schema_filepath": "/home/airflow/gcs/dags/schemas/",
-  "sentry_dsn": "https://9e0a056541c3445083329b072f2df690@o14203.ingest.sentry.io/6190849",
+  "sentry_dsn": "https://94027cdcc4c9470f9dafa2c0b456c2c9@o14203.ingest.sentry.io/5806618",
```
sydneynotthecity, line 172: Can we revert to the original DSN for the dev file? The `9e0a...` DSN points to our test project in Sentry.

sydneynotthecity (approved):
Looks good, I'd like to make sure we test this in `test-hubble319619` prior to releasing the code to prod so we are certain we are not breaking any DAGs.

lucaszanotelli:
I'll deploy this as it is to the `test-hubble` composer and merge it after testing it doesn't break anything.

PR #120, opened 2023-03-10
Add delete partition tasks to eho

sydneynotthecity:
Add a new task to dag definition: `delete_enrich_op_task` which will drop old partitions before inserting new records in the `enriched_history_operations` tables


chowbao (approved):
Linting changes make this much harder to read lol
But looks good to me

PR #127, opened 2023-04-05
Add dbt dags and task executor

chowbao:
https://stellarorg.atlassian.net/browse/HUBBLE-26
In conjunction with https://github.com/stellar/stellar-dbt/pull/22

Create airflow dags and task executor for dbt models.

New DAGs are created based on the same split of marts models in stellar-dbt:
* marts_tables (general marts tables/models with no subdirectory)
* enriched_tables
* partnership_assets
* ledger_current_state

Some mart tables have dependencies on other mart tables. There is no trigger dependency on a dependent table's completed run because the DAGs may run on different cadences.

The dbt profile is created from defined airflow variables:
* dbt_target
* dbt_dataset
* dbt_maximum_bytes_billed
* dbt_job_execution_timeout_seconds
* dbt_job_retries
* dbt_project
* dbt_threads
* dbt_private_key_id
* dbt_private_key
* dbt_client_email
* dbt_client_id
* dbt_auth_uri
* dbt_token_uri
* dbt_auth_provider_x509_cert_url
* dbt_client_x509_cert_url
* dbt_full_refresh_models

`dbt_full_refresh_models` should be a json that contains <model_name>: True/False to determine which model_names should run dbt with the flag `--full-refresh`. Model_names not in dbt_full_refresh_models will not run with the flag.

`dbt_job_retries`, `dbt_job_execution_timeout_seconds`, `dbt_threads` are separate from the airflow retry and concurrency configs

Service account credentials (`private_key_id`, `private_key`, etc...) uses the same values from the credentials.json files from the gcp project service account (same as the cloud functions credentials.json).
These secret values are stored in [gcp's secret manager](https://console.cloud.google.com/security/secret-manager?project=test-hubble-319619). Where the `airflow-variables-<var_name>` pattern (e.g. `airflow-variables-dbt_private_key` creates the airflow variable `dbt_private_key` that can be used in dags

TODO: update prod airflow variables in a followup PR when released


sydneynotthecity:
> dbt_job_retries, dbt_job_execution_timeout_seconds, dbt_threads are separate from the airflow retry and concurrency configs

Do you know how `dbt_job_retries` differs? If we set the `dbt_job_retries` to 2, for instance, would the Airflow DAG task run with both retries prior to failing? So then dbt is really retrying the code `dbt_job_retires` * `airflow_retries` times?

chowbao:
>> dbt_job_retries, dbt_job_execution_timeout_seconds, dbt_threads are separate from the airflow retry and concurrency configs

>Do you know how dbt_job_retries differs? If we set the dbt_job_retries to 2, for instance, would the Airflow DAG task run with both retries prior to failing? So then dbt is really retrying the code dbt_job_retires * airflow_retries times?

I haven't tested it but my assumption was yes it would do `dbt_job_retires * airflow_retries`. I think it makes sense to just set dbt_job_retries = 1 and then the actual retries we want to do in airflow retries

dags/enriched_tables_dag.py
```diff
@@ -0,0 +1,27 @@
...
+    default_args=get_default_dag_args(),
+    start_date=datetime.datetime(2023, 4, 4, 0, 0),
+    description='This DAG runs dbt to create the tables for the models in marts/enriched/.',
+    schedule_interval='0 */1 * * *', # Runs hourly
+    params={},
+    catchup=False,
+)
+
+# tasks for enriched tables
+enriched_history_operations = build_dbt_task(dag, 'enriched_history_operations')
```
sydneynotthecity, line 22: In the original pipeline, the enriched tables depends on history transactions, ledgers and operations to finish loading. Ledgers runs on a quick cadence than transactions/operations (every 15 min instead of every 30 min). We'll need to verify that we're not dropping records in the incremental loads based on the batch run date filter.

dags/ledger_current_state_dag.py
```diff
@@ -0,0 +1,34 @@
...
+from airflow.models.variable import Variable
+
+init_sentry()
+
+dag = DAG(
+    'ledger_current_state',
+    default_args=get_default_dag_args(),
+    start_date=datetime.datetime(2023, 4, 4, 0, 0),
+    description='This DAG runs dbt to create the tables for the models in marts/ledger_current_state/',
+    schedule_interval='0 */1 * * *', # Runs hourly
```
sydneynotthecity, line 16: these views take a while to run; if the execution time is garbage, we can up cadence to daily

dags/partnership_assets_dag.py
```diff
@@ -0,0 +1,42 @@
...
+init_sentry()
+
+dag = DAG(
+    'partnership_assets',
+    default_args=get_default_dag_args(),
+    start_date=datetime.datetime(2023, 4, 4, 0, 0),
+    description='This DAG runs dbt to create the partnership asset intermediate tables and aggregate tables.',
+    schedule_interval='0 11 * * *', # Daily 10 AM UTC
+    params={},
+    catchup=False,
```
sydneynotthecity, line 18: this one i would actually like to backfill 365 days of data so we can have trending for market cap. Doesn't need to be in this PR, but as an FYI.

dags/stellar_etl_airflow/build_dbt_task.py
```diff
@@ -0,0 +1,103 @@
...
+         dag=dag,
+         do_xcom_push=True,
+         is_delete_operator_pod=True,
+         startup_timeout_seconds=720,
+         in_cluster=in_cluster,
+         config_file=config_file_location,
+         affinity=affinity,
+         container_resources=k8s.V1ResourceRequirements(requests=resources_requests),
+         on_failure_callback=alert_after_max_retries,
+         image_pull_policy='Always'
```
sydneynotthecity, line 102: would love to figure out if there is a way for us to update this in the future to `ifNotPresent`

sydneynotthecity (approved):
Everything looks good to continue testing. There is a way to add cross dag dependencies. It's a little goofy, but we do it [here](https://github.com/stellar/stellar-etl-airflow/blob/master/dags/stellar_etl_airflow/build_cross_dependency_task.py) before [loading](https://github.com/stellar/stellar-etl-airflow/blob/master/dags/history_archive_with_captive_core_dag.py#L242-L244) the enriched_history_operations table. 

I don't think it's necessary for us to add those dependencies in the new loading mechanism of enriched history operations table; just a thought.

dags/partnership_assets_dag.py
chowbao: Added the correct catchup date. In theory airflow would run for the 365 period. If it's too slow we can do the backfill manually

dags/stellar_etl_airflow/build_dbt_task.py
chowbao: Added TODO to code and created JIRA ticket

sydneynotthecity (approved):
Updates look good

PR #137, opened 2023-04-18
dag created for mgi insertion in bigquery

cayod:
Dag was created to automate the insertion of new data in the Bigquery from Gcs, in this Dag we have a sensor that checks if has a file that was inserted on the date of the run, when this file is found it will send this data to Bigquery on mgi table.


dags/mgi_pipeline_dag.py
```diff
@@ -0,0 +1,46 @@
...
+internal_dataset = Variable.get("bq_dataset")
+public_project = Variable.get("public_project")
+public_dataset = Variable.get("public_dataset")
+bucket_name = Variable.get("bucket_mgi")
+
+
+check_gcs_file = build_check_gcs_file('check_gcs_file', dag, bucket_name)
+
+
+send_mgi_to_bq_internal_task = build_gcs_csv_to_bq_task('send_mgi_to_bq_pub_task', dag, internal_project, internal_dataset, 'mgi', bucket_name)
```
sydneynotthecity, line 41: could we name this table `raw_mgi_stellar_transactions`?

schemas/mgi_schema.json
```diff
@@ -0,0 +1,92 @@
...
+        "name": "rec_face_tran_crncy_amt",
+        "type": "FLOAT"
+    },
+    {
+        "mode": "NULLABLE",
+        "name": "send_partner_name",
+        "type": "STRING"
+    },
+    {
+        "mode": "NULLABLE",
```
sydneynotthecity, line 68: missing: `receive_partner_name`

sydneynotthecity (commented):
We're missing one field from the schema. The rest are nits

sydneynotthecity (approved):
thanks for the changes, looks good!

PR #143, opened 2023-04-26
Fix state tables and dbt

sydneynotthecity:
PR fixes a couple of bugs in Airflow:

1. the `airflow_variables_dev.json` was missing the `bucket_mgi` variable, which is causing the production CI deploy to fail.
2. State tables that use the `export_ledger_entry_changes` task needed to use the old stellar etl command. That code does not print stats about successful and failed transforms, so the tasks were failing with the new code.
3. the dbt macro that passed Airflow execution date needed the date set in an env variable, `EXECUTION_DATE` so that the code used the correct run date. Since no variable was passed, partnership asset code was originally defaulting every run to current date, which broke backfilling.


dags/ddls/queries/v_liquidity_pool_value.sql
```diff
@@ -0,0 +1,80 @@
...
+        from `hubble-261722.crypto_stellar_internal_2.v_liquidity_pools_current` as lp
+        where deleted is false
+    )
+    , asset_xlm_price as (
+        select
+            asset_code
+            , asset_issuer
+            , price_in_xlm
+            , rank() over (partition by asset_code, asset_issuer order by last_updated_ts desc) as rank_nr
+        from `hubble-261722.crypto_stellar_internal_2.asset_prices_xlm`
```
chowbao, line 23: Nit: should we move away from hard coding tables?

chowbao (approved):

PR #162, opened 2023-06-08
New Airflow Task: Copy Tables for current ledger state

sydneynotthecity:
The current ledger state tables are only available in the internal Hubble dataset. The public dataset also needs access to these tables. There were several options regarding how to make this data available:

1. Create views that pointed to the underlying internal tables,
2. Copy the tables to the public dataset after loading internal
3. Duplicate the dbt jobs to load the tables directly from `stellar-dbt`

I chose not to use views due to limitations of data accessibility. Users would not have access to the underlying table metadata (clustering, partitioning), they could not export the data via BigQuery API, which would limit the ways in which user could interact with the data. Duplicating dbt jobs seemed confusing and messy, we don't want to maintain two copies of the same job, and there should be a single source of truth for these tables.

I opted to build a new Airflow task, `build_copy_task`, that uses the `BigQueryInsertJobOperator` to copy the data from the internal tables to new destination tables in the public dataset. The tasks execute after the tables finish loading from dbt.

### DIFF
--- a/dags/ledger_current_state_dag.py
+++ b/dags/ledger_current_state_dag.py
@@ -2,6 +2,7 @@ import datetime
 
 from airflow import DAG
 from airflow.models.variable import Variable
+from stellar_etl_airflow.build_copy_table_task import build_copy_table
 from stellar_etl_airflow.build_dbt_task import build_dbt_task
 from stellar_etl_airflow.default import get_default_dag_args, init_sentry
 
@@ -17,6 +18,11 @@ dag = DAG(
     catchup=False,
 )
 
+internal_project = Variable.get("bq_project")
+internal_dataset = Variable.get("bq_project")
+public_project = Variable.get("public_project")
+public_dataset = Variable.get("public_dataset")
+
 # tasks for staging tables for ledger_current_state tables
 stg_account_signers = build_dbt_task(dag, "stg_account_signers")
 stg_history_ledgers = build_dbt_task(dag, "stg_history_ledgers")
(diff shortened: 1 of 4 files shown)


dags/stellar_etl_airflow/build_copy_table_task.py
```diff
@@ -0,0 +1,63 @@
...
+            },
+            "createDisposition": "CREATE_IF_NEEDED",
+            "writeDisposition": write_disposition,
+        }
+    }
+
+    return BigQueryInsertJobOperator(
+        task_id=f"copy_table_{source_table}",
+        execution_timeout=timedelta(
+            seconds=Variable.get("task_timeout", deserialize_json=True)[
```
chowbao, line 56: Need to add to airflow vars for test to pass

dags/ledger_current_state_dag.py
```diff
@@ -26,11 +32,63 @@
...
+accounts_current = build_dbt_task(dag, accounts)
+liquidity_pools = "liquidity_pools_current"
+liquidity_pools_current = build_dbt_task(dag, liquidity_pools)
+offers = "offers_current"
+offers_current = build_dbt_task(dag, offers)
+trust_lines = "trust_lines_current"
+trust_lines_current = build_dbt_task(dag, trust_lines)
+
+# copy the tables over to public dataset
+account_signers_public = build_copy_table(
```
chowbao, line 47: Should all these write_disposition be `WRITE_APPEND ` or `WRITE_TRUNCATE`? 

It defaults to `WRITE_APPEND` but I think it should be `WRITE_TRUNCATE` so that both datasets will have the same data at all times

dags/stellar_etl_airflow/build_copy_table_task.py
sydneynotthecity: Good catch

dags/ledger_current_state_dag.py
sydneynotthecity: you're right--idk why I set to `WRITE_APPEND`. `WRITE_TRUNCATE` is the desired disposition for these tables

chowbao (approved):

PR #174, opened 2023-06-30
Canvas sandbox dag

cayod:
This PR is a dag that updates canvas sandbox monthly, a dataset name canvas_sandbox that has 6 months of transactional tables, state tables, and views that point to the current state tables.

Tables:
accounts
history_assets
claimable_balances
history_effects
history_ledgers
liquidity_pools
offers
history_operations
account_signers
history_trades
history_transactions
trust_lines
enriched_history_operations

Views:
v_account_signers_current
v_accounts_current
v_claimable_balances_current
v_liquidity_pool_trade_volume
v_liquidity_pool_value
v_liquidity_pools_current
v_liquidity_providers
v_offers_current
v_trust_lines_current


dags/canva_sandbox_dag.py
sydneynotthecity, line 1: nit: can we update the dag name to `canvas_sandbox_dag`?

dags/canva_sandbox_dag.py
```diff
@@ -0,0 +1,216 @@
...
+    user_defined_filters={"fromjson": lambda s: json.loads(s)},
+    catchup=False,
+)
+
+table_ids = Variable.get("table_ids", deserialize_json=True)
+view_ids = Variable.get("view_ids", deserialize_json=True)
+project = Variable.get("bq_project")
+dataset = Variable.get("bq_dataset")
+sandbox_dataset = Variable.get("sandbox_dataset")
+test_project_check = "test-hubble-319619"
```
sydneynotthecity, line 36: Would this be better as an airflow variable rather than hard coded in the DAG?

dags/canva_sandbox_dag.py
```diff
@@ -0,0 +1,216 @@
...
+check_view_liquidity_pool_value = path_to_execute(dag, "check_for_test")
+if project != test_project_check:
+    update_view_liquidity_pool_value = build_bq_insert_job_sandbox(
+        dag,
+        project,
+        dataset,
+        view_ids["v_liquidity_pool_value"],
+        sandbox_dataset,
+        view=True,
+    )
```
sydneynotthecity, line 123: Why is only the liquidity pool value view updated if the project is in test? Can you give some more context regarding this design choice?

dags/stellar_etl_airflow/build_bq_insert_job_sandbox_task.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def build_bq_insert_job_sandbox(
+    dag, project, dataset, table, sandbox_dataset, view=False
+):
+    UPDATE_TABLES = (
+        f"CREATE OR REPLACE TABLE {sandbox_dataset}.{table} AS ( "
+        f"SELECT * "
+        f"FROM {dataset}.{table} "
+        f"WHERE batch_run_date >= DATE_SUB(CURRENT_DATE(), INTERVAL 6 MONTH))"
```
sydneynotthecity, line 40: Do you think any of these tables are big enough where it would be useful to partition or cluster them? Since we have smaller grain, we could even partition by day instead of month if that makes sense.

sydneynotthecity (commented):
Overall the change is starting to come together. Couple feedback items:

- The tests are failing in the CI/CD pipeline and need to be fixed 
- I would like to delete the `v_[table]_current` views in the production project because we now [build tables](https://github.com/stellar/stellar-dbt/tree/master/models/marts/ledger_current_state) in dbt that present the current network view. Can you adapt your code to instead create a view that is `SELECT * FROM [table]_current`?

dags/canva_sandbox_dag.py
cayod: The v_liquidity_pool_value view doesn't exist in dev, so when we run in dev, this task will be empty and don't generate some error

dags/stellar_etl_airflow/build_bq_insert_job_sandbox_task.py
cayod: It's a good idea, i will do it

dags/canva_sandbox_dag.py
sydneynotthecity: ah, ok that makes sense

dags/queries/create_table.sql
```diff
@@ -0,0 +1,7 @@
+create or replace table `{project_id}.{target_dataset}.{table_id}`
+partition by date_trunc(batch_run_date, month)
+options (partition_expiration_days = 180) as (
```
sydneynotthecity, line 3: awesome, I like this design where the partitions automatically expire :)

dags/sandbox_create_dag.py
```diff
@@ -0,0 +1,88 @@
...
+)
+
+init_sentry()
+
+with DAG(
+    "sandbox_create_dag",
+    default_args=get_default_dag_args(),
+    start_date=datetime.datetime(2023, 1, 1),
+    description="This DAG creates a sandbox",
+    schedule_interval="0 0 1 * *",
```
sydneynotthecity, line 28: let's push this back to 6am UTC so that we get a full picture of the prior day

dags/sandbox_update_dag.py
```diff
@@ -0,0 +1,61 @@
...
+)
+
+init_sentry()
+
+with DAG(
+    "sandbox_update_dag",
+    default_args=get_default_dag_args(),
+    start_date=datetime.datetime(2023, 1, 1),
+    description="This DAG updates a sandbox",
+    schedule_interval="0 0 1 * *",
```
sydneynotthecity, line 28: move back to 6am

dags/sandbox_create_dag.py
sydneynotthecity: actually, should this [schedule](https://airflow.apache.org/docs/apache-airflow/1.10.5/scheduler.html) be `@once` and then we re-trigger only when the tables/views need to be recreated because of a schema change? curious your thoughts on that since we will be incrementally updating the tables with the `sandbox_update_dag`

sydneynotthecity (approved):
I have nothing that blocks this DAG, just some small suggestions. Feel free to merge once the `sandbox_dataset` is `crypto_stellar_internal_sandbox`.

PR #176, opened 2023-07-10
dag partners mgi

cayod:
This PR updates all the partner's tables following the same pattern of Mgi CSV, for adding a new partner in the dag, we only need to add a schema file in the schema folder and the prefix_folder, prefix_id, and table in the partners_data variable. The dag will search for the data in the partners_bucket variable inside the prefix_folder folder, looking for the CSV with the prefix_id variable and today's date.


dags/partner_pipeline_dag.py
sydneynotthecity, line 1: Will this DAG replace the current MGI pipeline dag?

dags/partner_pipeline_dag.py
```diff
@@ -0,0 +1,78 @@
...
+            skip_leading_rows=1,
+            schema_fields=read_local_schema(PARTNERS[partner]["table"]),
+            write_disposition="WRITE_TRUNCATE",
+            dag=dag,
+            retry_delay=timedelta(minutes=5),
+            max_retry_delay=timedelta(minutes=30),
+            on_failure_callback=alert_after_max_retries,
+        )
+
+        start_tables_task >> check_gcs_file >> send_partner_to_bq_internal_task
```
sydneynotthecity, line 78: curious if we are going to run into the same issues that we're seeing with the orchestration for canvas update dag. If you make changes to those DAGs, you might need to propogate those changes here as well.

sydneynotthecity (approved):
DAG looks good. I'm curious if the partner pipeline dag is meant to replace the current MGI pipeline dag. If it is we'll need to delete the MGI dag once this is released as cleanup

dags/partner_pipeline_dag.py
cayod: Yes, we only need to know if you agree with this folder organization and define what will be that name of the bucket in Hubble that will get all the partner's data

sydneynotthecity: Yes, I like this folder organization. I just created a bucket `ext-partner-sftp` for these files

cayod: Ok, the partners_bucket variable was updated for ext-partner-sftp

sydneynotthecity:
Ops request to update file location of sftp job: https://github.com/stellar/ops/issues/2413

PR #186, opened 2023-07-28
Update raw mgi schema

sydneynotthecity:
MGI is adding address fields to the raw file, as detailed in [JIRA-183](https://stellarorg.atlassian.net/browse/MGI-183)

Raw schema needs to be updated so that the file change does not break the pipeline. Fields will be added to downstream tables later.


Shaptic (approved):
I looked at the CSV file attached in [MGI-183](https://stellarorg.atlassian.net/browse/MGI-183) and made sure each of these `partnr_` columns was reflected in this schema. I dunno how to check data types but they all seem fine, too!

ire-and-curses (approved):

PR #188, opened 2023-08-07
Update ledger closed

cayod:
This PR update the schema with new field closed_at in tables: `account`, `account_signers`, `claimabla_balances`, `history_assets`, `history_effects`, `history_operations`, `history_transactions`, `liquidity_pools`, `offers`, and `trust_lines`. Also, update the partition from tables using `closed_at` to `batch_run_date`.

### DIFF
--- a/airflow_variables.json
+++ b/airflow_variables.json
@@ -173,19 +173,19 @@
     },
     "accounts": {
       "type": "MONTH",
-      "field": "closed_at"
+      "field": "batch_run_date"
     },
     "account_signers": {
       "type": "MONTH",
-      "field": "closed_at"
+      "field": "batch_run_date"
     },
     "claimable_balances": {
       "type": "MONTH",
-      "field": "closed_at"
+      "field": "batch_run_date"
     },
     "history_effects": {
       "type": "MONTH",
-      "field": "closed_at"
+      "field": "batch_run_date"
     },
     "history_ledgers": {
       "type": "MONTH",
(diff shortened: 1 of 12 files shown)


airflow_variables.json
sydneynotthecity, line 1: We actually want the _opposite_. The tables partitioned by `batch_run_date` should be repartitioned by `closed_at`. Using `closed_at` as the partition date is more intuitive for the end users.

schemas/history_assets_schema.json
```diff
@@ -38,5 +38,10 @@
     "mode": "NULLABLE",
     "name": "batch_insert_ts",
     "type": "TIMESTAMP"
+  },
+  {
+    "mode": "NULLABLE",
+    "name": "closed_at",
+    "type": "TIMESTAMP"
```
sydneynotthecity, line 45: Can be dropped from `history_assets` based on our conversation at standup.

sydneynotthecity (commented):
Tables should actually be partitioned by `closed_at` instead of `batch_run_date`. Once the changes have been made, I will approve. This PR should not be merged until the corresponding stellar etl [PR](https://github.com/stellar/stellar-etl/pull/187) is merged and new image is deployed

sydneynotthecity (approved):
Looks good!

sydneynotthecity:
@cayod and @chowbao I think we should separate this into two separate releases:

1. Add `ledger_closed_at` to the tables
2. Repartition the tables using the `ledger_closed_at` instead of `batch_run_date`

The reason why I think this should be two separate releases is we have a lot of scheduled queries, dbt code, and reports that query by date. We should identify these queries and coordinate deploying changes in the queries with repartitioning the tables, otherwise we will impact the downstream code that executes on top of the tables.

PR #189, opened 2023-08-08
fix: fixed schema upload in cicd

cayod:
This PR fixes the directory that schema folders that are uploaded


sydneynotthecity (approved):
LGTM

PR #207, opened 2023-08-31
update dbt variable

cayod:
This PR updates the Dbt image with the last image that inserts closed_at field in history_transactional, operations, and effects.


sydneynotthecity (commented):
dbt image has updated, the latest tag is `fcbee51`

sydneynotthecity (approved):

PR #212, opened 2023-09-08
Soroban support testnet

chowbao:
Diff of soroban changes with master


dags/state_table_dag.py
```diff
@@ -114,6 +121,18 @@
...
+)
+delete_contract_code_pub_new_task = build_delete_data_task(
+    dag, public_project, public_dataset_new, table_names["contract_code"]
+)
+delete_config_settings_pub_new_task = build_delete_data_task(
+    dag, public_project, public_dataset_new, table_names["config_settings"]
+)
+delete_expiration_pub_new_task = build_delete_data_task(
+    dag, public_project, public_dataset_new, table_names["expiration"]
+)
```
sydneynotthecity, line 135: For the new contract data tasks, can they be renamed to `delete_contract_data_task`, `delete_contract_code_task`, etc? 

Since we are not supporting multiple copies of the tables, there is no reason to differentiate the task name.

sydneynotthecity (approved):
LGTM

PR #213, opened 2023-09-14
insert alert_after_max_retries in cross_dependency and check_executions_date task

cayod:
This PR implements the alert_after_max_retries in tasks that don't have this parameter.


sydneynotthecity (approved):
LGTM

PR #214, opened 2023-09-20
update dbt image

cayod:
This PR updates an image from Dbt in dev and prod.

### DIFF
--- a/airflow_variables.json
+++ b/airflow_variables.json
@@ -112,7 +112,7 @@
     "partnership_assets__asset_activity_fact": false,
     "trade_agg": false
   },
-  "dbt_image_name": "stellar/stellar-dbt:2a44526",
+  "dbt_image_name": "stellar/stellar-dbt:0643adc",
   "dbt_job_execution_timeout_seconds": 6000,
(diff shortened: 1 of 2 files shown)


sydneynotthecity (approved):
LGTM. I wonder, for the purposes of this repo does it make since to pin the `dbt_image_name` to `stellar/stellar-dbt:latest`? 

We can manually update the image name in airflow, and I don't know how useful it is to keep the versioned history of that image tag in this repo. @chowbao thoughts?

cayod:
> LGTM. I wonder, for the purposes of this repo does it make since to pin the `dbt_image_name` to `stellar/stellar-dbt:latest`?
> 
> We can manually update the image name in airflow, and I don't know how useful it is to keep the versioned history of that image tag in this repo. @chowbao thoughts?

The Kubernetes website inserted a recomendation about image pull in kubelet:
You should avoid using the :latest tag when deploying containers in production as it is harder to track which version of the image is running and more difficult to roll back properly.
https://kubernetes.io/docs/concepts/containers/images/

PR #218, opened 2023-09-20
Soroban support testnet

chowbao:
Soroban airflow changes


airflow_variables.json
```diff
@@ -91,7 +91,11 @@
       "transaction_id",
       "account",
       "type"
-    ]
+    ],
+    "contract_data": ["last_modified_ledger"],
+    "contract_code": ["last_modified_ledger"],
+    "config_settings": ["last_modified_ledger"],
+    "expiration": ["last_modified_ledger"]
```
sydneynotthecity, line 98: something to think about before the next testnet reset: should these fields be clustered by `contract_id` as well? Config settings is fine as-is. I can create a ticket if we want to change the table structure.

dags/stellar_etl_airflow/build_export_task.py
```diff
@@ -39,7 +41,12 @@ def select_correct_filename(cmd_type, base_name, batched_name):
 
 
 def generate_etl_cmd(
-    command, base_filename, cmd_type, use_gcs=False, use_testnet=False
+    command,
+    base_filename,
+    cmd_type,
+    use_gcs=False,
```
sydneynotthecity, line 47: Uhhhhh could we be changing the way we write out files to GCS and just haven't? 🤦‍♀️ 

Don't change now, but adding a ticket to the new ETL overhaul epic to address.

PR #219, opened 2023-09-21
Public dbt dag

cayod:
This PR implements a dag for public Dbt para as tabelas: 
history_assets
enriched_history_operations

The commented lines in the dag are the next tables that will be inserted.

### DIFF
--- a/dags/public_marts_tables_dag.py
+++ b/dags/public_marts_tables_dag.py
@@ -19,6 +19,7 @@ dag = DAG(
 )
 
 start_dag = DummyOperator(task_id="start_dag", dag=dag)
+
 start_dag
 
 env = Variable.get("environment")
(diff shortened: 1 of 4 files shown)


dags/public_marts_tables_dag.py
sydneynotthecity, line 1: Is this DAG meant to only execute in production? How will we be able to stage/test this code when there are changes? I'm also curious if it makes more sense to change the `project` based on `env` passed instead of defining the entire DAG based on the environment.

sydneynotthecity (commented):
Overall looks good. I'd like to clarify how we will test updates to this pipeline before approving it.

dags/public_marts_tables_dag.py
cayod: I updated the task to be able to test the code.

sydneynotthecity (approved):
Thanks for the changes, looks good

PR #220, opened 2023-09-29
change target project from canvas sandbox to dev-hubble

cayod:
This PR updates the canvas sandbox project target to dev-hubble


sydneynotthecity (commented):
Will this update only change the canvas sandbox to `dev-hubble`? I think we want to keep the canvas sandbox in prod, and everything else data eng related that we own should move to `dev-hubble`. The reason being is that the canvas sandbox is production data and should be viewed as a production-like env. Whereas dev hubble is more of an ephemeral local dev environment with no guarantees

PR #226, opened 2023-10-06
fixed auto update from airflow variables

cayod:
This PR fixes the airflow variables update that wasn't happening in ci and release


cayod:
For the new changes to work in ci and release, we need to grant the service account that deploys the dags the following permissions:
`composer.environments.create`
`composer.environments.executeAirflowCommand`
`composer.environments.update`

The service accounts are:
`test-airflow-cicd@test-hubble-319619.iam.gserviceaccount.com`
`airflow-cicd@hubble-261722.iam.gserviceaccount.com`

cayod:
The CI failed, I was wrong about this permission. Can we grant `Environment and Storage Object User` role to this service account? @sydneynotthecity

sydneynotthecity:
@cayod I updated to grant permissions to the appropriate buckets. Can you test again?

cayod:
I was wrong about this last role. Can we grant the `Composer Environment Admin` role to this service account? This is the right role for uploading the variables in composer. @sydneynotthecity

sydneynotthecity:
@cayod updated the permissions again

.github/workflows/release.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          cat airflow_variables.json | sed -e s/\\/home\\/airflow\\/gcs\\/dags\\/// > airflow_variables_ci.json
+          cat airflow_variables_prod.json | sed -e s/\\/home\\/airflow\\/gcs\\/dags\\/// > airflow_variables_ci.json
```
sydneynotthecity, line 24: @cayod do we not need the same line in the `ci.yml` for the test release?

.github/workflows/release.yml
```diff
@@ -65,3 +65,16 @@ jobs:
         env:
           GOOGLE_CLOUD_PROJECT: hubble-261722
           BUCKET: us-central1-hubble-2-d948d67b-bucket
+
+      - name: Update Airflow variables
+        uses: actions-hub/gcloud@master
+        env:
+          PROJECT_ID: hubble-2-d948d67b
```
sydneynotthecity, line 72: i don't think this project id is correct. Looks like the part of the composer bucket name

dags/stellar_etl_airflow/add_files_to_composer.py
```diff
@@ -53,7 +55,11 @@ def upload_dags_to_composer(bucket_name: str, env: str) -> None:
 
         for f in files:
             # remove path to temp dir
-            if f.endswith(".json"):
+            if f.endswith("variables.json"):
```
sydneynotthecity, line 58: Neither filename will end with `*variables.json` if you are appending the environment name. This might need to be updated to `if f.endwith(f"variables_{env}.json")

dags/stellar_etl_airflow/add_files_to_composer.py
```diff
@@ -26,7 +26,9 @@ def _create_files_list(env: str) -> Tuple[str, List[str]]:
     # copy airflow configuration file
     copy(f"airflow-{env}.cfg", f"{temp_dir}/airflow.cfg")
     copy(f"airflow-{env}.cfg", f"airflow.cfg")
-
+    # copy airflow variables json
+    copy(f"airflow_variables_{env}.json", f"{temp_dir}/variables.json")
+    copy(f"airflow_variables_{env}.json", f"variables.json")
```
sydneynotthecity, line 31: The prod file name is `airflow_variables.json` and does not contain `prod` in the name. Do we need to update?

sydneynotthecity (commented):
Couple of questions to make sure the release behavior is the same in production as test.

dags/stellar_etl_airflow/add_files_to_composer.py
cayod: I already updated the airflow_variables.json to airflow_variables_prod.json to insert this change in the code.

.github/workflows/release.yml
cayod: Sorry @sydneynotthecity, I didn't understand this question.

sydneynotthecity: Checked the master file and we're good. Idk why the diffs aren't showing up accurately here

sydneynotthecity (approved):

PR #242, opened 2023-11-03
Add `asset_id` to `history_assets` table

lucaszanotelli:
This PR adds the missing field `asset_id` in the query that deduplicates the `history_assets` table.


sydneynotthecity (approved):
LGTM

PR #244, opened 2023-11-07
Update airflow vars for dbt-image

chowbao:


airflow_variables_prod.json
```diff
@@ -1,6 +1,22 @@
 {
   "affinity": {
-    "default": {},
+    "default": {
```
sydneynotthecity, line 3: Can you give me an overview of what this change will do?

sydneynotthecity (approved):

airflow_variables_prod.json
chowbao: For some reason the blank `default` was causing the mart dag to fail 🤷‍♀️ 
In theory it should be a noop where the jobs will run in the default pool node cluster anyways

PR #254, opened 2023-11-24
Change `dbt` authentication method

lucaszanotelli:
This PR simplifies the authentication method used to connect `dbt` to BigQuery and delete the former used variables.

The `oauth` method will use the authentication from the service account of the GKE node pool running the `dbt` operations.

The former authentication method was exposing the whole JSON service account in the Airflow UI.


airflow_variables_prod.json
```diff
@@ -119,16 +114,12 @@
...
   "dbt_mart_dataset": "sdf_marts",
   "dbt_maximum_bytes_billed": 100000000000000,
-  "dbt_private_key": "",
-  "dbt_private_key_id": "",
   "dbt_project": "hubble-261722",
   "dbt_target": "prod",
   "dbt_threads": 1,
-  "dbt_token_uri": "https://oauth2.googleapis.com/token",
-  "gcs_exported_data_bucket_name": "us-central1-hubble-2-d948d67b-bucket",
+  "gcs_exported_data_abucket_name": "us-central1-hubble-2-d948d67b-bucket",
```
sydneynotthecity, line 122: nit: I think the variable name is wrong, did you mean `gcs_exported_data_bucket_name`?

sydneynotthecity (approved):
LGTM

Do you know if we need to enable any new permissions on the GKE service account in order for `oauth` to work?

PR #255, opened 2023-11-24
Add timestamp field to mgi raw table

cayod:
This PR adds a timestamp field to the `raw_mgi_stellar_transactions` table, this field will be used in the task for data quality checks with the Elementary.


dags/partner_pipeline_dag.py
```diff
@@ -75,4 +79,24 @@
             on_failure_callback=alert_after_max_retries,
         )
 
-        start_tables_task >> check_gcs_file >> send_partner_to_bq_internal_task
+        insert_ts_field = BigQueryInsertJobOperator(
```
lucaszanotelli, line 82: Instead of creating a new job, couldn't we just set a default value for new this column? Like this https://cloud.google.com/bigquery/docs/default-values

schemas/raw_mgi_stellar_transactions_schema.json
```diff
@@ -198,5 +198,10 @@
     "mode": "NULLABLE",
     "name": "cancellation_reas",
     "type": "STRING"
+  },
+  {
+    "mode": "NULLABLE",
+    "name": "timestamp",
```
lucaszanotelli, line 204: It's better rename this to `loaded_at` or something else. `timestamp` might be a reserved name and could lead to errors.

sydneynotthecity: ➕ agree with Lucas. if you want to stay consistent with our base tables, we use `batch_insert_ts` as the column name

sydneynotthecity (approved):
Will you need me to alter the production table to add the same default?

cayod:
> Will you need me to alter the production table to add the same default?

I will try here and let you know if needed

PR #260, opened 2023-11-29
Fix JSON Jinja templates

lucaszanotelli:
This PR fixes type error from output of Jinja template and rollback some variables were being used on top-level code which make impossible to use Jinja templates.


sydneynotthecity (approved):
LGTM

PR #264, opened 2023-12-01
Revert getting network variables from template

lucaszanotelli:
This PR undo getting network variables from template since it didn't parse the values correctly in production.


sydneynotthecity (approved):
Thanks!

PR #272, opened 2023-12-12
Add dag for update stellar-base.min.js

cayod:
With the need to automate `stellar-base.min.js` to use as a package for Dbt, we created a dag to automatically download this library.

This dag does a check to verify the file's upload date in GCS and checks the date of the last tag that is not beta or alpha in the repository. If this date is greater than that of GCS, the file is downloaded and replaced in GCS.

### DIFF
--- a/airflow_variables_dev.json
+++ b/airflow_variables_dev.json
@@ -299,5 +299,9 @@
   },
   "currency_bucket": "currencies_ohlc",
   "max_db_entry_age_in_days": 90,
-  "public_dbt_image_name": "stellar/stellar-dbt:dbf348d"
+  "public_dbt_image_name": "stellar/stellar-dbt:dbf348d",
+  "stellar_base_js_repositoy_name": "bower-js-stellar-base",
+  "stellar_base_js_repository_url": "https://raw.githubusercontent.com/stellar/bower-js-stellar-base/master/stellar-base.min.js",
+  "package_js_bucket_name": "stellar-etl-cli",
+  "package_js_object_name": "stellar-base.min.js"
 }
--- a/dags/js_library_update_dag.py
+++ b/dags/js_library_update_dag.py
@@ -1,0 +1,70 @@
+from datetime import datetime
+from json import loads
+
+from airflow import DAG
+from airflow.operators.empty import EmptyOperator
+from airflow.operators.python import BranchPythonOperator
+from airflow.operators.python_operator import PythonOperator
+from stellar_etl_airflow.build_check_latest_tag_task import is_new_tag
+from stellar_etl_airflow.build_get_update_library_task import (
+    check_file_upload_date,
+    send_library_to_gcs,
+)
+from stellar_etl_airflow.default import get_default_dag_args, init_sentry
+
+init_sentry()
+
+dag = DAG(
+    "js_library_update_dag",
+    default_args=get_default_dag_args(),
+    start_date=datetime(2015, 9, 30),
+    description="This DAG updates the js library from bower-js-stellar-base repo.",
+    schedule_interval="0 17 * * *",  # Daily 11 AM UTC
+    render_template_as_native_obj=True,
+    params={
+        "alias": "js_library",
+    },
+    user_defined_filters={"fromjson": lambda s: loads(s)},
+    catchup=True,
+    max_active_runs=1,
+)
+
+repo = "{{ var.value.stellar_base_js_repositoy_name }}"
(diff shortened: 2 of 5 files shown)


dags/stellar_etl_airflow/build_check_latest_tag_task.py
```diff
@@ -0,0 +1,55 @@
...
+    latest_tag["commit_date"] = latest_tag["commit_date"].split("T")[0]
+    last_known_tag_date = str(last_known_tag_date)
+    if latest_tag:
+        logging.info(f"Latest tag date: {latest_tag['commit_date']}")
+        if latest_tag["commit_date"] > last_known_tag_date:
+            logging.info("New tag found!")
+            return "update_library_task"
+        else:
+            logging.info("No new tags found")
+            return "task_aready_up_to_date_task"
```
lucaszanotelli, line 55: You misspelled `already`.

dags/js_library_update_dag.py
```diff
@@ -0,0 +1,69 @@
...
+    op_kwargs={
+        "file_url": file_url,
+        "bucket_name": bucket_name,
+        "destination_blob_name": destination_blob_name,
+    },
+    dag=dag,
+)
+
+task_aready_up_to_date = EmptyOperator(
+    task_id="task_aready_up_to_date_task",
```
lucaszanotelli, line 65: You misspelled `already`.

airflow_variables_dev.json
```diff
@@ -299,5 +299,9 @@
   },
   "currency_bucket": "currencies_ohlc",
   "max_db_entry_age_in_days": 90,
-  "public_dbt_image_name": "stellar/stellar-dbt:dbf348d"
+  "public_dbt_image_name": "stellar/stellar-dbt:dbf348d",
+  "stellar_base_js_repositoy_name": "bower-js-stellar-base",
+  "stellar_base_js_repository_url": "https://raw.githubusercontent.com/stellar/bower-js-stellar-base/master/stellar-base.min.js",
+  "package_js_bucket_name": "stellar-etl-cli",
+  "package_js_object_name": "stellar-base.min.js"
```
lucaszanotelli, line 306: Add those variables to `airflow_variables_prod.json` also.

dags/js_library_update_dag.py
```diff
@@ -0,0 +1,69 @@
...
+from stellar_etl_airflow.default import get_default_dag_args, init_sentry
+
+init_sentry()
+
+dag = DAG(
+    "js_library_update_dag",
+    default_args=get_default_dag_args(),
+    start_date=datetime(2023, 1, 1),
+    description="This DAG updates the js library from bower-js-stellar-base repo.",
+    schedule_interval="0 17 * * *",  # Daily 11 AM UTC
```
lucaszanotelli, line 22: Change the comment to represent the actual schedule.

lucaszanotelli:
This DAG should be paused in the `test-hubble` Airflow. You can achieve that by mimicking the `dags/dataset_reset_dag.py`.

sydneynotthecity:
Related context: This work was part of a project to research the feasibility of parsing raw XDR within dbt using SQL UDFs. We wanted to explore whether we could materialize raw XDR in a table and create all downstream data structures through dbt models.

We have determine it is feasible, however we believe this architecture is more confusing and hard to support versus maintaining the `stellar-etl` library. We would be dependent on the js-sdk and would have less control parsing the XDR than we have today.

As part of the work, we did cost analysis between executing stellar-etl vs dbt models:

> By analyzing the data from the DAG executed with XDR data compared to the dag that extracts data from history_transaction, we were able to verify that the DAG with XDR data has an average time of 6 minutes and 52 minutes more than the normal history_transaction flow, with an average cost from US$22.42 per month with virtual machine costs taking into account the resources used to perform the tasks. To execute the DBT query that transforms the XDR data into the history_transaction table with an average duration of 2 minutes and 11 seconds and use of 776500 Slots in milliseconds, we were able to calculate an average cost of $3.15 per month taking into account the cost of $0.06 per slot hour that Bigquery has in the [Enterprise edition](https://cloud.google.com/bigquery/pricing#bigquery-pricing).
> For summarize:
> A DBT query costs $0.002184 x 1440 (number of runs every half hour for a month) = $3.14 per month.
> Extra dag execution time 412 sec (06:52) for each run = [$22.42 per month](https://cloud.google.com/products/calculator/#id=96edad27-1764-4558-a236-f2535462eee5).

sydneynotthecity:
Closing the PR until the need arises to update the stellar-base js sdk

PR #273, opened 2023-12-12
Remove elementary model from `dbt run`

lucaszanotelli:
This PR removes the `elementary` model from the `dbt run` (`build_dbt_task()`) so that it doesn't truncate the tables.

This also changes the default dbt command from `run` to `build` that is more complete.


sydneynotthecity (approved):
LGTM

PR #277, opened 2023-12-22
Add public dbt DAGs

lucaszanotelli:
This PR add 3 DAGs that are only triggered when the `enriched_history_operations` finishes.

To give the right pace to each DAG, some tasks were added to the `history_archive_with_captive_core_dag`:

- `enriched_history_operations_dag` will be trigger **every 30 minutes**.
- `dbt_daily_dag` will be triggered **once a day** after the EHO loads.
- `dbt_ohlc_dag` will be triggered **twice a day** after the EHO loads.

This also introduces a new way to build a task that runs `dbt` that is more flexible and clear.
**Important: merge the [PR#110 from stellar/stellar-dbt](https://github.com/stellar/stellar-dbt/pull/110) and get the new image tag and substitute the techindicium/stellar-dbt first.**


dags/history_archive_with_captive_core_dag.py
```diff
@@ -353,6 +356,60 @@
...
+    trigger_dag_id="dbt_daily_dag",
+    execution_date="{{ ts }}",
+    reset_dag_run=True,
+    wait_for_completion=True,
+    dag=dag,
+)
+trigger_dbt_ohlc_dag = TriggerDagRunOperator(
+    task_id="trigger_dbt_ohlc_dag",
+    trigger_dag_id="dbt_ohlc_dag",
+    trigger_rule="none_failed_min_one_success",
```
sydneynotthecity, line 403: does this mean the task will only trigger if the prior run succeeded?

sydneynotthecity: and there are no failed tasks in the dag?

dags/stellar_etl_airflow/build_dbt_task.py
```diff
@@ -57,8 +57,85 @@ def create_dbt_profile(project="prod"):
...
+    logging.info(f"sh commands to run in pod: {args}")
+
+    return KubernetesPodOperator(
+        task_id=f"dbt_{command_type}_{task_name}",
+        name=f"dbt_{command_type}_{task_name}",
+        namespace=Variable.get("k8s_namespace"),
+        service_account_name=Variable.get("k8s_service_account"),
+        env_vars={
+            "DBT_USE_COLORS": "0",
+            "DBT_DATASET": "crypto_stellar_dbt",
```
chowbao, line 114: Can the mart tables be added to the existing `crypto_stellar` dataset instead of creating a new dataset specifically for dbt?

dags/stellar_etl_airflow/build_dbt_task.py
```diff
@@ -57,8 +57,85 @@ def create_dbt_profile(project="prod"):
...
+        config_file=config_file_location,
+        affinity=affinity,
+        container_resources=container_resources,
+        on_failure_callback=alert_after_max_retries,
+        image_pull_policy="IfNotPresent",
+        image_pull_secrets=[k8s.V1LocalObjectReference("private-docker-auth-indicium")],
+    )
+
+
 def build_dbt_task(
```
chowbao, line 137: Should the function names `build_dbt_task` and `dbt_run_task` be flipped?
It looks like the `dbt_run_task` function calls `dbt build` and `build_dbt_task` calls `dbt run`

dags/dbt_daily_dag.py
```diff
@@ -0,0 +1,24 @@
...
+    catchup=False,
+    tags=["dbt"],
+)
+
+dbt_daily_task = dbt_run_task(
+    dag,
+    flag="exclude",
+    operator="",
+    model_name="enriched_history_operations",
+    tag="ohlc",
```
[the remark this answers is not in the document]
lucaszanotelli: This DAG triggers every model but the `enriched_history_operations` model and the ones that contains the `ohlc` tag. It's simpler than adding a `daily` tag to a bunch of models. This way we can benefit from the `dbt` data lineage when running the tool.

dags/history_archive_with_captive_core_dag.py
```diff
@@ -353,6 +356,60 @@
...
+"""
+These tasks use the BranchDateTimeOperator to trigger different dbt DAGs daily and every 12 hours
+"""
+last_dag_run_task = BranchDateTimeOperator(
+    task_id="last_dag_run_task",
+    use_task_logical_date=True,
+    follow_task_ids_if_true=["trigger_dbt_daily_dag"],
+    follow_task_ids_if_false=["not_last_dag_run_task"],
+    target_upper=time(17, 30, 1),
+    target_lower=time(17, 30, 0),
```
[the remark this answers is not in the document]
lucaszanotelli: This time interval was only for testing. I forgot to change. I'll also add a comment to describe better what's the intent behind these tasks.

dags/history_archive_with_captive_core_dag.py
lucaszanotelli: I'll add a description to make it more clear.

dags/stellar_etl_airflow/build_dbt_task.py
lucaszanotelli: For now this DAG will run in parallel with the other `dbt` DAGs. The idea is to replace — after everything is validated — the current dataset (`crypto_stellar`) by the new one (`crypto_stellar_dbt`).

dags/stellar_etl_airflow/build_dbt_task.py
lucaszanotelli: The idea is to maintain the `build_dbt_task` while validating so let's keep it for now. But I agree that `dbt_run_task` is confusing. I'll rename it to a more generic name, since this is the only task that uses `dbt`.

airflow_variables_dev.json
```diff
@@ -338,5 +338,5 @@
   },
   "currency_bucket": "currencies_ohlc",
   "max_db_entry_age_in_days": 90,
-  "public_dbt_image_name": "stellar/stellar-dbt:dbf348d"
+  "public_dbt_image_name": "stellar/stellar-dbt:346a2d0"
```
sydneynotthecity, line 341: This commit hash actually doesn't build an image, you might want to consider updating to `ddca412 `, which was the last successful build, or wait for [this PR ](https://github.com/stellar/stellar-dbt/pull/112)to merge

sydneynotthecity (approved):
Warning: the dbt image you tagged did not actually build in Jenkins correctly. I left a comment with the latest successful build tag.

airflow_variables_dev.json
lucaszanotelli: I'll update to `ddca412` for now just to see the dbt running in the test environment.

dags/history_archive_with_captive_core_dag.py
```diff
@@ -294,6 +297,76 @@
     cluster=True,
 )
 
+"""
+This task triggers the `enriched_history_operations_dag` that's resposible for running the dbt `enriched_history_operations` model.
+The execution date of the triggered DAG will be the same as this DAG.
+The `reset_dag_run` boolean set to `True` means that whenever this DAG is rerun/cleared, the triggered DAG also receives a new DAG run.
+The `wait_for_completion` boolean set to `True` means that this task will wait until the triggered DAG finishes.
+"""
+trigger_eho_dag = TriggerDagRunOperator(
```
chowbao, line 306: Is it possible to change these trigger dags in `history_archive_with_captive_core_dag` to be instead wait triggers in the eho, daily, and ohlc dags?

This seems to cause the unintended issue of when `history_archive_with_captive_core_dag` can't trigger eho, daily, or ohlc that `history_archive_with_captive_core_dag` will be stuck in a retry loop. This can cause the DAG to fall behind until the issue eho, daily, or ohlc is fixed

airflow_variables_dev.json
```diff
@@ -338,5 +338,5 @@
   },
   "currency_bucket": "currencies_ohlc",
   "max_db_entry_age_in_days": 90,
-  "public_dbt_image_name": "stellar/stellar-dbt:dbf348d"
+  "public_dbt_image_name": "stellar/stellar-dbt:a480828"
```
chowbao, line 341: Oh `stellar/stellar-etl:a480828` was for the "image_name" not for the dbt image

dags/history_archive_with_captive_core_dag.py
lucaszanotelli: Set `wait_for_completion` isn't really an option because the other pipelines depend on eho completion. I'll decouple the logic from the `history_archive_with_captive_core` DAG and propose another DAG to run `dbt`.

PR #282, opened 2024-01-16
Hot Fix: Audit log Dag

sydneynotthecity:
The insert job in BQ should only take the `schemaUpdateOptions` parameter if the write disposition is `WRITE_APPEND`. Moved the parameter to be optionally added to the query config.

### DIFF
--- a/dags/stellar_etl_airflow/build_bq_insert_job_task.py
+++ b/dags/stellar_etl_airflow/build_bq_insert_job_task.py
@@ -74,7 +74,7 @@ def build_bq_insert_job(
     if create:
         configuration["query"]["createDisposition"] = "CREATE_IF_NEEDED"
     if write_disposition == "WRITE_APPEND":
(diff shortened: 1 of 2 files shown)


chowbao (approved):

dags/stellar_etl_airflow/build_bq_insert_job_task.py
```diff
@@ -74,6 +73,8 @@ def build_bq_insert_job(
         configuration["query"]["clustering"] = {"fields": cluster_fields[table]}
     if create:
         configuration["query"]["createDisposition"] = "CREATE_IF_NEEDED"
+    if write_disposition == "WRITE_APPEND":
+        configuration["query"]["schemaUpdateOptions"] = "ALLOW_FIELD_ADDITION"
```
chowbao, line 77: Oh I think this is supposed to be a list

sydneynotthecity: you're right. Though weirdly the DAG tasks still complete successfully in test. I'll still make the update though

PR #285, opened 2024-01-18
Add empty task to make OHLC task triggerable midday

lucaszanotelli:
This PR adds a task in between `trigger_dbt_ohlc_dag` and `midday_run_task` so the task `trigger_dbt_ohlc_dag` can be triggered by the success run of the `trigger_dbt_daily_dag` task also.


sydneynotthecity (approved):
LGTM

PR #289, opened 2024-01-22
Update dbt image and params

chowbao:


sydneynotthecity (approved):
LGTM

Let's make sure to validate everything in test before releasing this change to the release branch

dags/dbt_daily_dag.py
```diff
@@ -14,19 +14,54 @@
 dag = DAG(
     "dbt_daily",
```
chowbao, line 15: Need better DAG names

dags/stellar_etl_airflow/build_dbt_task.py
```diff
@@ -56,7 +56,6 @@ def create_dbt_profile(project="prod"):
 
     return create_dbt_profile_cmd
 
-
 def dbt_task(
```
chowbao, line 59: We should refactor this function. Probably in a follow up PR

airflow_variables_dev.json
```diff
@@ -117,19 +117,19 @@
...
-  "dbt_dataset": "test_sdf",
+  "dbt_dataset": "test",
   "dbt_full_refresh_models": {
-    "partnership_assets__account_holders_activity_fact": true,
-    "partnership_assets__asset_activity_fact": true
+    "partnership_assets__account_holders_activity_fact": false,
+    "partnership_assets__asset_activity_fact": false
   },
-  "dbt_image_name": "stellar/stellar-dbt:738ac3e",
+  "dbt_image_name": "chowbao/stellar-dbt:test8",
```
chowbao, line 125: Need to update image

dags/stellar_etl_airflow/build_dbt_task.py
sydneynotthecity: Agreed. Let's not block this PR, but `dbt_task` and `build_dbt_task` should be consolidated into one function. There is no need to have both

dags/dbt_daily_dag.py
```diff
@@ -1,32 +1,45 @@
...
 from stellar_etl_airflow.build_dbt_task import dbt_task
 from stellar_etl_airflow.default import get_default_dag_args
 
 dag = DAG(
     "dbt_daily",
     default_args=get_default_dag_args(),
-    schedule_interval=None,  # don’t schedule since this DAG is externally triggered daily
+    start_date=datetime(2024, 1, 23, 0, 0),
+    description="This DAG runs dbt models at a daily cadence",
+    schedule_interval="0 13 * * *",  # Runs at 13:00 UTC
```
sydneynotthecity, line 14: i think it's ok for us to kick this off at 00:00, or closer to the start of a new day. Is there a benefit to running the middle of the day?

dags/dbt_daily_dag.py
```diff
@@ -1,32 +1,45 @@
...
+# DBT models to run
+ohlc_task = dbt_task(dag, tag="ohlc")
+liquidity_pool_trade_volume_task = dbt_task(dag, tag="liquidity_pool_trade_volume")
+mgi_task = dbt_task(dag, tag="mgi")
+liquidity_providers_task = dbt_task(dag, tag="liquidity_providers")
+trade_agg_task = dbt_task(dag, tag="trade_agg")
+fee_stats_agg_task = dbt_task(dag, tag="fee_stats_agg")
+asset_stats_agg_task = dbt_task(dag, tag="asset_stats_agg")
+network_stats_agg_task = dbt_task(dag, tag="network_stats_agg")
+
```
sydneynotthecity, line 37: i think we're missing the partnership assets tag

dags/dbt_half_hourly_dag.py
```diff
@@ -0,0 +1,38 @@
...
+    max_active_runs=3,
+    catchup=True,
+    tags=["dbt-half-hourly"],
+)
+
+# Wait on ingestion DAGs
+wait_on_cc = build_cross_deps(
+    dag, "wait_on_ledgers_txs", "history_archive_without_captive_core"
+)
+wait_on_state_table = build_cross_deps(dag, "wait_on_state_table", "state_table_export")
```
sydneynotthecity, line 29: we're missing a third dependency on the new history_archive_with_captive_core_combined. The `wait_on_cc` task right now will only wait for `history_ledgers` table to complete

dags/stellar_etl_airflow/build_dbt_task.py
```diff
@@ -82,7 +82,7 @@ def dbt_task(
     )
     affinity = Variable.get("affinity", deserialize_json=True).get(resource_cfg)
 
-    dbt_image = "{{ var.value.public_dbt_image_name }}"
+    dbt_image = "{{ var.value.dbt_image_name }}"
```
sydneynotthecity, line 85: todo: we should delete the public dbt image name var in general once we're confident we don't need it

sydneynotthecity (commented):
I'll help brainstorm better dag names tomorrow. We're missing a tag, otherwise everything looks good

dags/dbt_daily_dag.py
chowbao: Oh I did this because 13 UTC is 8 AM. So if anything did go wrong we would see it first thing in the morning 

00:00 UTC is 7 PM EST 🤮

dags/dbt_daily_dag.py
chowbao: Ah good catch

dags/dbt_half_hourly_dag.py
chowbao: Ah that's right

dags/dbt_enriched_base_tables_dag.py
```diff
@@ -0,0 +1,44 @@
...
+    "dbt_enriched_base_tables",
+    default_args=get_default_dag_args(),
+    start_date=datetime(2024, 1, 26, 0, 0),
+    description="This DAG runs dbt models at a half hourly cadence",
+    schedule_interval="*/30 * * * *",  # Runs every 30 mins
+    user_defined_filters={
+        "container_resources": lambda s: k8s.V1ResourceRequirements(requests=s),
+    },
+    max_active_runs=3,
+    catchup=True,
```
sydneynotthecity, line 21: One thing we may need to watch for: the current_state tables don't actually need to catchup if ever this DAG is unscheduled for some reason. Backfilling for enriched tables will need to be a separate process apart from `current_state` to avoid putting unnecessary load on the db

sydneynotthecity (approved):
Looks good

PR #292, opened 2024-01-29
Updates to get dbt models running in prod

chowbao:
Update/Add variables to get dbt models working in prod

### DIFF
--- a/dags/partner_pipeline_dag.py
+++ b/dags/partner_pipeline_dag.py
@@ -24,7 +24,7 @@ with DAG(
     default_args=get_default_dag_args(),
     start_date=datetime(2023, 1, 1, 0, 0),
     description="This DAG automates daily updates to partner tables in BigQuery.",
-    schedule_interval="20 15 * * *",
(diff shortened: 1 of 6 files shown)


sydneynotthecity (approved):
Looks good

dags/partner_pipeline_dag.py
```diff
@@ -24,7 +24,7 @@
     default_args=get_default_dag_args(),
     start_date=datetime(2023, 1, 1, 0, 0),
     description="This DAG automates daily updates to partner tables in BigQuery.",
-    schedule_interval="20 15 * * *",
+    schedule_interval="0 13 * * *",
```
sydneynotthecity, line 27: note: MGI does not deliver the file until 16:00 UTC, so we'll be polling for a bit before the file shows up

chowbao: That should be fine. Seems like we were waiting then anyways lol.

Also the way the cross dependency function is written makes it weird cause it's hard coded to accept a minutes before offset

PR #301, opened 2024-02-20
Unify dbt sdf marts in one dbt task

cayod:
This PR adds an if statement that allows you to insert more than one tag for the `dbt_task` function, with this we can run all the tags in the same run to prevent errors about concurrent updates mentioned in the HUBBLE-301 ticket.

### DIFF
--- a/airflow_variables_dev.json
+++ b/airflow_variables_dev.json
@@ -354,7 +354,8 @@
       "asset_stats",
       "network_stats",
       "partnership_assets",
-      "history_assets"
+      "history_assets",
+      "soroban"
     ]
   }
 }
--- a/airflow_variables_dev.json
+++ b/airflow_variables_dev.json
@@ -342,5 +342,19 @@
   "internal_source_db": "test-hubble-319619",
   "internal_source_schema": "test_crypto_stellar_internal",
   "public_source_db": "test-hubble-319619",
-  "public_source_schema": "test_crypto_stellar"
+  "public_source_schema": "test_crypto_stellar",
+  "dbt_tags": {
+    "sdf_marts": [
+      "ohlc",
+      "liquidity_pool_trade_volume",
+      "mgi",
+      "liquidity_providers",
+      "trade_agg",
+      "fee_stats",
+      "asset_stats",
+      "network_stats",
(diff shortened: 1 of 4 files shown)


chowbao:
As discussed in the meeting I think we would like to keep each tag running in a separate task. The elementary issue was more pronounced because we were running everything as a separate task (stg -> conform -> mart etc...) so there were tons of tasks finishing and writing to elementary at the same time causing the error.

Where did you see this causing the elementary issue? Do you have link to an example in airflow?

cayod:
> As discussed in the meeting I think we would like to keep each tag running in a separate task. The elementary issue was more pronounced because we were running everything as a separate task (stg -> conform -> mart etc...) so there were tons of tasks finishing and writing to elementary at the same time causing the error.
> 
> Where did you see this causing the elementary issue? Do you have link to an example in airflow?

Before I had not seen this error anywhere except in the ticket you had opened, I created this PR because I realized that this sdf_marts Dag could suffer from this problem due to the separate tasks. However, today while I was creating a task for Airflow that sends alerts from elementary to slack I received this error while running Dag sdf_marts.

https://he02f27a661269b05p-tp.appspot.com/log?dag_id=dbt_sdf_marts&task_id=dbt_build_partnership_assets&execution_date=2024-02-21T16%3A00%3A00%2B00%3A00

chowbao:
> Before I had not seen this error anywhere except in the ticket you had opened, I created this PR because I realized that this sdf_marts Dag could suffer from this problem due to the separate tasks. However, today while I was creating a task for Airflow that sends alerts from elementary to slack I received this error while running Dag sdf_marts. look at the first attempt at this task. https://he02f27a661269b05p-tp.appspot.com/log?dag_id=dbt_sdf_marts&task_id=dbt_build_partnership_assets&execution_date=2024-02-21T16%3A00%3A00%2B00%3A00

I think the frequency of the issues is low enough where it should be fine for it to happen and we just let the task retry. If we condense all models to be run at the same time it makes debugging much harder. 
The other option is to run the tags serially instead of in parallel but I'm not sure that's necessary either.

I would say we can probably close this issue

PR #302, opened 2024-02-20
Update dbt image

chowbao:
* Update dbt image
* Enable alerts in dbt_sdf_marts


dags/dbt_sdf_marts_dag.py
```diff
@@ -6,6 +6,8 @@
 from stellar_etl_airflow.build_dbt_task import dbt_task
 from stellar_etl_airflow.default import get_default_dag_args
 
+init_sentry()
```
sydneynotthecity, line 9: smart

sydneynotthecity (approved):
LGTM

dags/dbt_sdf_marts_dag.py
chowbao: Much good

PR #304, opened 2024-02-22
Update export task params

chowbao:
stellar-etl params updated to use the common export interface


sydneynotthecity (approved):
exciting! make sure test isn't borked before promoting to prod since we've had some test issues this week.

PR #306, opened 2024-02-22
Delete unused dags

chowbao:
Delete unused dags


sydneynotthecity (approved):
LGTM.

As discussed, our CI/CD pipeline does not delete production content, so you will need to manually delete DAGs from prod.

PR #327, opened 2024-03-12
Add new fee columns to history transactions

chowbao:


schemas/history_transactions_schema.json
```diff
@@ -178,5 +178,20 @@
     "mode": "NULLABLE",
     "name": "transaction_result_code",
     "type": "STRING"
+  },
+  {
+    "mode": "NULLABLE",
+    "name": "inclusion_fee_bid",
+    "type": "STRING"
```
sydneynotthecity, line 185: should these be int?

sydneynotthecity (commented):
I think there is a copy/pasta error in the new column definitions. Lmk if they are intentionally defined as string

schemas/history_transactions_schema.json
chowbao: Oh oops you're right

sydneynotthecity (approved):
LGTM

PR #333, opened 2024-03-24
Integrate cdp changes to ingest from datastore txmeta files

chowbao:
Integrate cdp changes to ingest from datastore txmeta files
* Add cdp_* airflow parameters to run in parallel with captive core mode. cdp_* dags should write to cdp_* dataset
* cdp_* versions of the captive core export dags
* New parameters to select between captive core mode and reading txmeta mode in the dags and in build_export_task
* build time task did not need to be adjusted. Current stellar-etl implementation pulls from history archives which should be fine no matter what mode is being run


chowbao:
Oops this was supposed to be draft PR 🤷‍♀️

dags/cdp_state_table_dag.py
sydneynotthecity, line 1: Do you think it always makes sense to keep this DAG separate from the `history_*` DAG? Or any merit in consolidating into a single source ingestion dag?

dags/stellar_etl_airflow/build_export_task.py
```diff
@@ -47,6 +47,8 @@ def generate_etl_cmd(
     use_gcs=False,
     use_testnet=False,
     use_futurenet=False,
+    use_captive_core=False,
+    txmeta_datastore_url="",
```
sydneynotthecity, line 51: Does this mean that by default, `build_export_task` will fail because a datastore url is not passed?

sydneynotthecity (approved):
What are your thoughts on how we're going to deploy cdp and replace the old DAGs? Curious how you're thinking about that.

dags/stellar_etl_airflow/build_export_task.py
chowbao: Yeah it'll fail. My thought was that when `build_export_task` is called it should have one explicitly set to be able to run

dags/cdp_history_archive_with_captive_core_combined_export_dag.py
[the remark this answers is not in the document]
chowbao: Merged history_* dags into a single `history_tables_dag`. As discussed we'll have a state table and history table dag for now as we figure out a better way to split up our dags

airflow_variables_prod.json
```diff
@@ -363,5 +363,7 @@
   "public_source_schema": "crypto_stellar",
   "slack_elementary_channel": "alerts-hubble-data-quality",
   "elementary_secret": "slack-token-elementary",
-  "dbt_elementary_dataset": "elementary"
+  "dbt_elementary_dataset": "elementary",
+  "use_captive_core": "False",
+  "txmeta_datastore_url": "gcs://ledger-exporter/ledgers/testnet"
```
sydneynotthecity, line 368: should this be `mainnet`?

dags/history_tables_dag.py
```diff
@@ -22,12 +22,12 @@
...
 dag = DAG(
-    "history_archive_with_captive_core",
+    "history_table_export",
     default_args=get_default_dag_args(),
-    start_date=datetime(2023, 12, 18, 17, 0),
+    start_date=datetime(2024, 4, 22, 19, 0),
     catchup=True,
     description="This DAG exports trades and operations from the history archive using CaptiveCore. This supports parsing sponsorship and AMMs.",
-    schedule_interval="*/30 * * * *",
+    schedule_interval="*/10 * * * *",
```
sydneynotthecity, line 30: Any data/benchmarking yet on how well this DAG does on a 10 min interval?

In prod, we'll need to release in a couple steps to make sure we don't mess up the schedule interval shift:

- update old DAGs to include end date
- delete old DAGs
- add new DAG with new start date (that won't miss any gaps)

dags/history_tables_dag.py
```diff
@@ -126,6 +138,32 @@
...
+)
+asset_export_task = build_export_task(
+    dag,
+    "archive",
+    "export_assets",
+    "{{ var.json.output_file_names.assets }}",
+    use_testnet=use_testnet,
+    use_futurenet=use_futurenet,
+    use_gcs=True,
+    resource_cfg="wocc",
```
sydneynotthecity, line 164: do we actually use the k8s resource_cfg with composer 2?

sydneynotthecity (approved):
LGTM. Nothing blocking the merge

airflow_variables_prod.json
chowbao: Ah it should be pubnet

dags/history_tables_dag.py
chowbao: In test it takes about ~10 minutes to run (mostly cause of the slow startup time issue)

dags/history_tables_dag.py
chowbao: We don't. This is just a noop

PR #336, opened 2024-04-02
Add successful transforms to analyze rows

laysabit:
For the sake of counting the rows to analyze any duplicated row, when comparing batches, we must use the successful transforms variable, which tell us how many lines were transferred. When it is working as part of the output value in XCOM, another task can be created to analyze it.


sydneynotthecity (approved):
LGTM

PR #350, opened 2024-05-01
Update Canvas sandbox dependencies

sydneynotthecity:
The `sandbox_update` DAG broke because we changed the name of the upstream DAG that the workflow was dependent on. This PR addresses the name change and refactors the code to be less brittle.

- The DAG originally updated data based on `current_date()`. If the pipeline broke, backfilling missing data required manual intervention. The query is now date parameterized by `batch_start_date` and will backfill missing days because `catchup=True`.
- copy statements pointed to the internal dataset, `hubble-261722.crypto_stellar_internal_2`. Updated the statements to use `crypto-stellar.crypto_stellar`.
- Added the state table DAG as an additional dependency. We missed this

_Note:_ We will probably need to run the `sandbox_create` DAG because the column order on the public tables differ from internal. It'll be easier to just recreate the tables so that the schemas match.

Future considerations:

- add data quality checks 
- use a merge statement to ensure duplicate data is not inserted into the table (alternative is to delete/reinsert)
- utilize [table clone ](https://cloud.google.com/bigquery/docs/table-clones-intro)feature to copy data instead of manually inserting new data via SQL

### DIFF
--- a/dags/queries/create_table.sql
+++ b/dags/queries/create_table.sql
@@ -3,6 +3,7 @@ partition by date_trunc(batch_run_date, day)
 options (partition_expiration_days = 180) as (
     select *
     from `{project_id}.{dataset_id}.{table_id}`
-    where batch_run_date >= date_sub(current_date(), interval 6 month)
-    and batch_run_date < date_sub(current_date())
+    where
+        batch_run_date >= date_sub(current_date(), interval 6 month)
+        and batch_run_date < date_sub(current_date())
 )
--- a/dags/queries/create_table.sql
+++ b/dags/queries/create_table.sql
@@ -5,5 +5,5 @@ options (partition_expiration_days = 180) as (
     from `{project_id}.{dataset_id}.{table_id}`
     where
         batch_run_date >= date_trunc(date_sub(current_date(), interval 6 month), day)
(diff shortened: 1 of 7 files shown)


dags/queries/create_table.sql
```diff
@@ -4,4 +4,5 @@ options (partition_expiration_days = 180) as (
     select *
     from `{project_id}.{dataset_id}.{table_id}`
     where batch_run_date >= date_sub(current_date(), interval 6 month)
+    and batch_run_date < date_sub(current_date())
```
chowbao, line 7:
```suggestion
    and batch_run_date < current_date()
```

dags/queries/update_table.sql
```diff
@@ -1,4 +1,4 @@
 insert into {project_id}.{target_dataset}.{table_id}
 select *
 from {project_id}.{dataset_id}.{table_id}
-where date_trunc(batch_run_date, day) = date_trunc(current_date() - interval 1 day, day)
+where date_trunc(batch_run_date, day) = date_trunc({batch_run_date})
```
chowbao, line 4:
```suggestion
where date_trunc(batch_run_date, day) = date_trunc({batch_run_date}, day)
```

chowbao (approved):

PR #356, opened 2024-05-07
Refactor dags to delete internal sources

laysabit:
Deleted from internal: account_signers, accounts, claimable_balances, history_effects, history_ledgers, history_operations, history_trades, history_transactions, trustlines, offers, liquidity_pools, history_assets.

Added new stellar-dbt image to dev and prod: "stellar/stellar-dbt:f360f91"

Along with it, it was deleted from bigQuery the views related to the sources that are not being used anymore
image

### DIFF
--- a/airflow_variables_dev.json
+++ b/airflow_variables_dev.json
@@ -52,7 +52,7 @@
     "partnership_assets__account_holders_activity_fact": false,
     "partnership_assets__asset_activity_fact": false
   },
-  "dbt_image_name": "stellar/stellar-dbt:6ee7342",
+  "dbt_image_name": "stellar/stellar-dbt:f360f91",
   "dbt_job_execution_timeout_seconds": 300,
   "dbt_job_retries": 1,
   "dbt_mart_dataset": "test_crypto_stellar_dbt",
--- a/airflow_variables_prod.json
+++ b/airflow_variables_prod.json
@@ -57,7 +57,7 @@
     "partnership_assets__asset_activity_fact": false,
     "trade_agg": false
   },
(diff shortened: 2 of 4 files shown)


laysabit:
the PR is ready for review! :) @sydneynotthecity

sydneynotthecity (commented):
Changes look good. @laysabit why are we keeping the `history_assets` table in the internal dataset? It looks like we deleted all other tables. I'm curious why we are keeping that one specifically.

laysabit:
> Changes look good. @laysabit why are we keeping the `history_assets` table in the internal dataset? It looks like we deleted all other tables. I'm curious why we are keeping that one specifically.

The sources are https://github.com/stellar/stellar-dbt/blob/master/models/sources/dev_sources.yml , correct? If those are the sources, assets does not categorize as a source, because it is not there, but if it is a source, I can remove it from the dag

sydneynotthecity:
Ah, I see the confusion. We actually are planning on building the `history_assets` table as a stellar-dbt-public [model](https://github.com/stellar/stellar-dbt-public/blob/master/models/marts/history_assets.sql). So it's actually a derived table that is a source table in downstream models. Please remove from the internal dataset.

laysabit:
> Ah, I see the confusion. We actually are planning on building the `history_assets` table as a stellar-dbt-public [model](https://github.com/stellar/stellar-dbt-public/blob/master/models/marts/history_assets.sql). So it's actually a derived table that is a source table in downstream models. Please remove from the internal dataset.

Done

sydneynotthecity (approved):
LGTM

PR #359, opened 2024-05-08
Feature/add slas

edualvess:
## Overview

This PR addresses the development of notification callbacks and the enabling of SLA checks for tasks that could potentially take longer to finish.

## Key Changes

- Defined `alert_sla_miss` function to send notifications to designed Slack channels leveraging Sentry integration
- Setup DAG-level parameter `sla_miss_callback` with the newly defined callback function for the required DAGs
- Setup appropriate SLA for each of the desired tasks
- Extra: Increased `task_timeout` value for the `get_ledger_range_from_times` in order to avoid task retries while processing XCOM


dags/stellar_etl_airflow/build_dbt_task.py
```diff
@@ -136,6 +136,9 @@ def dbt_task(
         on_failure_callback=alert_after_max_retries,
         image_pull_policy="IfNotPresent",
         image_pull_secrets=[k8s.V1LocalObjectReference("private-docker-auth")],
+        sla=timedelta(
+            seconds=Variable.get("task_sla", deserialize_json=True)[task_name]
+        ),
```
chowbao, line 141: This should be added to the function below, `build_dbt_task`, as well

airflow_variables_dev.json
```diff
@@ -245,7 +245,29 @@
     "build_delete_data_task": 180,
     "build_export_task": 180,
     "build_gcs_to_bq_task": 300,
-    "build_time_task": 120
+    "build_time_task": 300
+  },
+  "task_sla": {
+    "get_ledger_range_from_times": 240,
+    "export_task": 240,
```
chowbao, line 252: I think we need to adjust some of these `sla times`. Like for instance this `export_task` sla 240 will never be hit because the `task_timeout` for `export_task` is 180.

chowbao:
I believe you are missing the `sla=` parameter for some more the the build_* functions like for [BigQueryInsertJobOperator](https://github.com/stellar/stellar-etl-airflow/blob/master/dags/stellar_etl_airflow/build_bq_insert_job_task.py#L79).

I'd do a grep/search for all the operators/build functions in `dags/stellar_etl_airflow/` and see where `sla=` can be added. And then update the `task_sla` airflow variable. Or include a default value in `task_sla` that faster running tasks can be set to.

airflow_variables_dev.json
chowbao: Also the current sla times are too aggressive from all the alerts we are getting in the slack channel #alerts-hubble-testnet

edualvess: > I think we need to adjust some of these `sla times`. Like for instance this `export_task` sla 240 will never be hit because the `task_timeout` for `export_task` is 180.

There's an interesting behavior going on for the `export_task`. Even though the `task_timeout` is set for 180 seconds, the tasks usually run for around 220 seconds without triggering the timeout. Differently from `time_task`, where the timeout triggers in the middle of the job.

I'll push the commits where I increase the timeout for `export_task` too, which makes more sense to the actual behavior of the pipeline.

edualvess: > Also the current sla times are too aggressive from all the alerts we are getting in the slack channel #alerts-hubble-testnet

Those have been reviewed through the suggested buffer and will be updated

airflow_variables_prod.json
```diff
@@ -238,9 +238,33 @@
...
+    "fee_stats": 300,
+    "asset_stats": 360,
+    "network_stats": 300,
+    "partnership_assets": 1200,
+    "history_assets": 360,
+    "soroban": 360,
+    "snapshot_state": 840,
+    "elementary_dbt_sdf_marts": 120,
+    "build_bq_insert_job": 120,
+    "sandbox_create": 120
```
chowbao, line 267: I think the sandbox update `task_sla` is missing

dags/stellar_etl_airflow/build_time_task.py
```diff
@@ -53,6 +58,11 @@ def build_time_task(
         args.append("--testnet")
     elif use_futurenet:
         args.append("--futurenet")
+    # elif use_gcs:
+    #    args.extend(
+    #        ["--cloud-storage-bucket", Variable.get("gcs_exported_data_bucket_name")]
+    #    )
+    #    args.extend(["--cloud-provider", "gcp"])
```
chowbao, line 65: I think we should remove the code that is currently from the dev branch PR https://github.com/stellar/stellar-etl-airflow/pull/371

chowbao (approved):

PR #369, opened 2024-05-16
requirements updated

laysabit:
As I had to update prettier so the pre-commit can work, I took my time to also update sqlfluff to the version found in stellar-dbt.

Limited the quantity of the modules in the requirements to only the ones that are needed in the repository.


sydneynotthecity (approved):
LGTM

PR #377, opened 2024-06-04
HUBBLE-409 Update/delete airflow variable names

laysabit:
<details>

### PR Structure

      otherwise).

### Thoroughness

- [x] I've updated the README with the added features, breaking changes, new instructions on how to use the repository.

</details>

### What

README updated with new information.
Variable affinity is deleted.
Variable dbt_dataset was not deleted.

dbt_ was appended as prefix to these var names:

- internal_source_db
- internal_source_schema  
- public_source_db        
- public_source_schema    
- slack_elementary_channel
- elementary_secret     

### Why

dbt_dataset was not deleted because of its impact on the code:

how it should look without it:
how_should_look

how it looks without dbt_dataset:
how_it_looks

the error caused because of the lack of dbt_dataset:
error

### Known limitations

[N/A]


dags/dbt_stellar_marts_dag.py
```diff
@@ -20,7 +20,7 @@
     },
     max_active_runs=3,
     catchup=True,
-    tags=["dbt-sdf-marts"],
+    tags=["dbt-stellar-marts"],  # trocar
```
sydneynotthecity, line 23: nit: comment can be removed

dags/dbt_stellar_marts_dag.py
```diff
@@ -10,7 +10,7 @@
 init_sentry()
 
 dag = DAG(
-    "dbt_sdf_marts",
+    "dbt_stellar_marts",
```
sydneynotthecity, line 13: what was the reason for renaming this DAG?

sydneynotthecity (commented):
Looks good, thanks for cleaning up the airflow variables. One question needs addressed and then I'll approve the PR

dags/dbt_stellar_marts_dag.py
laysabit: it was a demand by Simon: "Rename dag dbt_sdf_marts to like dbt_stellar_marts as it’s a public repo and not only SDF marts" [Source](https://stellarorg.atlassian.net/jira/software/c/projects/HUBBLE/boards/65?search=409&selectedIssue=HUBBLE-409)

dags/dbt_stellar_marts_dag.py
laysabit: Comment is deleted

laysabit:
I changed the name of dbt_dataset, as it was asked by Eduardo Alves, to dbt_dataset_for_test, because it is only used in test environment, so it won't confuse the user.

sydneynotthecity (approved):
Thanks for the context, looks good!

PR #399, opened 2024-06-18
Add contract events to history dag

chowbao:
* Add history_contract_events to history dag
* Remove unused diagnostic_events task


airflow_variables_dev.json
sydneynotthecity, line 1: Do you mind also bumping the `dbt_image_name` to `1a777f9 `

schemas/history_contract_events_schema.json
```diff
@@ -0,0 +1,77 @@
...
+  },
+  {
+    "mode": "NULLABLE",
+    "name": "type_string",
+    "type": "STRING"
+  },
+  {
+    "mode": "NULLABLE",
+    "name": "topics",
+    "type": "JSON"
```
sydneynotthecity, line 30: ooh how did using the `JSON` data type work out?

dags/contract_events_test_dag.py
sydneynotthecity, line 1: out of curiosity, what was the reason for separating out `contract_events` into its own dag?

dags/contract_events_test_dag.py
```diff
@@ -0,0 +1,141 @@
...
+        "subtract_data_interval": macros.subtract_data_interval,
+        "batch_run_date_as_datetime_string": macros.batch_run_date_as_datetime_string,
+    },
+    sla_miss_callback=alert_sla_miss,
+)
+
+
+table_names = Variable.get("table_ids", deserialize_json=True)
+internal_project = "{{ var.value.bq_project }}"
+internal_dataset = "{{ var.value.bq_dataset }}"
```
sydneynotthecity, line 55: todo: clean up internal/public variables since we only write to one dataset

dags/contract_events_test_dag.py
chowbao: For testing purposes. I just wanted to get the `history_contract_events` table populated with some stuff without messing with the existing history dag

schemas/history_contract_events_schema.json
chowbao: It works out fine. I did have to wrap the `Topics` array into a dict

copying message from slack
> (https://stellarfoundation.slack.com/archives/C073MNCNVQQ/p1718755628622429)
test-hubble-319619.test_crypto_stellar.history_contract_events is up
I think it looks fine but the topics and topics_decoded json columns are a little wonky. You need to call them like select topics.topics or select topics_decoded.topics_decoded because they are normally []map[string]string (a list of dictionaries) but are now map[string][]map[string]string (a dict of list of dicts)

sydneynotthecity (approved):
Looks good, make sure to swap the image names once the new `stellar-etl` image is built

PR #413, opened 2024-06-27
[HUBBLE 444] Refactor Elementary monitoring to run every 30 min

edualvess:
<details>

### PR Structure

      otherwise).

### Thoroughness

</details>

### What

The `elementary_slack_alert_dbt_sdf_marts` task was moved into a separate DAG that executes independently of the dbt models after some data quality tests, as defined in [Alerting for dbt](https://docs.google.com/document/d/1ypdEsOFYmFx605GfWQ_IixDvGWddt5ZEeQ5K5c1S4z4/edit#heading=h.efewlepm4vgf) . This way, we will detect and trigger alerts on problems in a timely manner. The DAG is scheduled to run every 30 minutes, which aligns with the `dbt_enriched_base_tables` DAG.
The dbt tests running in the new DAG use a new tag designed to fit all the unit tests related to the models' data quality.

### Why

Elementary monitoring and alerting are currently executed at the very end of the dbt DAGs. If an upstream dbt model fails, this means that elementary never alerts to a data quality issue because the alerting depends on DAG execution status. The result is that alerts are late arriving or arrive after an issue has already been resolved.

### Known limitations

NA


dags/dbt_enriched_base_tables_dag.py
```diff
@@ -35,16 +34,12 @@
 wait_on_state_table = build_cross_deps(dag, "wait_on_state_table", "state_table_export")
 
 # DBT models to run
-enriched_history_operations_task = dbt_task(dag, tag="enriched_history_operations")
+enriched_history_operations_task = dbt_task(
+    dag, tag="enriched_history_operations", excluded="singular_test"
+)
 current_state_task = dbt_task(dag, tag="current_state")
```
chowbao, line 40: Does this also need `excluded="singular_test"`?

dags/dbt_stellar_marts_dag.py
```diff
@@ -55,8 +54,6 @@
 snapshot_state = dbt_task(dag, tag="snapshot_state")
 relevant_asset_trades = dbt_task(dag, tag="relevant_asset_trades")
```
chowbao, line 55: Same question for the dbt_tasks in here
Do they need `excluded="singular_test"` as well?

dags/dbt_enriched_base_tables_dag.py
edualvess: There's no need. I checked with the analytics team and this excluded tag for the singular test is related only to EHO, and the exclusion is mainly to avoid testing the same thing in two different workflows in a similar cadence.

dags/dbt_stellar_marts_dag.py
edualvess: Same response to the first comment, there's no need.

chowbao (approved):

PR #417, opened 2024-07-06
Update schemas

chowbao:
There is a new predicate depth added in one of the recent operations that caused errors in the airflow insert tasks
https://stellarfoundation.slack.com/archives/C02GQ9ZGNDU/p1720235944089819

Added the new `or` level to resolve this issue


sydneynotthecity:
> There is a new predicate depth added in one of the recent operations

Gross, if nothing else, we should convert `claimants` to JSON object instead of RECORD to avoid similar issues cropping up in two years

PR #419, opened 2024-07-08
[Hubble-446] Implement mechanism to only fail transient dbt task errors

edualvess:
<details>

### PR Structure

      otherwise).

### Thoroughness

- [x] I've updated the README with the added features, breaking changes, new instructions on how to use the repository.

</details>

### What
This PR addresses the develpment of a mechanism capable of controlling the **Airflow** task instances' flow in order to avoid unnecessary task retries.
As the key changes made to code:
- Developed `skip_retry_dbt_errors` function to avoid retrying non-transient dbt errors
- Added `skip_retry_dbt_errors` function as `on_retry_callback` for dbt tasks
- Added `dbt_transient_errors_patterns` variable to include the string patterns for known **dbt** transient errors

The approach followed for this development can be visualized in the flowchart:
transient

### Why

With the current dbt task setup, if a dbt test fails, Airflow will still retry the task many times. Without manual intervention, there will be no differing outcome for dbt tests, and we should not retry the task multiple times. Instead we would the dbt Airflow task to exhibit the following behavior:

Transient Errors - in the case of errors caused by Airflow infrastructure or transient issues, the task should continue to retry as originally designed. This is because we expect a different outcome for the task status with the retry.

Data Quality errors: if an error is triggered by bad data or a failed dbt test, the task should fail immediately if configured to fail. If the test is configured to warn, the task should continue successfully.


dags/stellar_etl_airflow/utils.py
```diff
@@ -0,0 +1,112 @@
...
+    logging.info("Task failed due to unforeseen error, proceeding to retry")
+    return True
+
+
+def read_log_file(log_file_path):
+    max_retries = 3
+    retry_delay = 30
+    log_contents = []
+
+    for attempt in range(max_retries):
```
sydneynotthecity, line 82: During testing, did you find that the log file sometimes didn't exist?

dags/stellar_etl_airflow/utils.py
```diff
@@ -0,0 +1,112 @@
...
+
+def skip_retry_dbt_errors(context) -> None:
+    """
+    Set task state to SKIPPED in case errors found in dbt are not transient.
+    """
+    if not check_dbt_transient_errors(context):
+        ti = context["ti"]
+        logging.info(
+            f"Set task instance {ti} state to \
+                {TaskInstanceState.SKIPPED} to skip retrying"
```
sydneynotthecity, line 107: To confirm, elementary will still surface this task failure as an alert since dbt will log a `FAILURE`, correct?

dags/stellar_etl_airflow/utils.py
```diff
@@ -0,0 +1,112 @@
...
+                            return True
+                logging.info("Could not find dbt transient errors, skipping retry")
+                return False
+            else:
+                logging.info(
+                    "dbt pipeline finished without errors, task failed but will not retry"
+                )
+                return False
+    # Check for transient errors in case dbt pipeline didn't finish
+    for line in log_contents:
```
sydneynotthecity, line 66: nit: This function loops through the `log_contents` twice. Do you think it's possible to move this logic up into the first for loop for `log_contents`? Not a big deal if not, these files are small.

sydneynotthecity (approved):
Changes look good, one small feedback and some questions for my own understanding, but none of it blocks the release.

dags/stellar_etl_airflow/utils.py
edualvess: Yes, this logic only runs after the first failure of the task in Airflow, be it due to dbt errors or any other. The elementary on-run-end hook have already finished running by the time this is called, so the alerts DAG should catch the warnings/failures results from the elementary dataset.

dags/stellar_etl_airflow/utils.py
edualvess: I did, but that was mostly my fault. I firstly developed this script to work at `on_execute_callback`, so I had to look for the log file related to the `ti.try_number`, but after migrating to the `on_retry_callback` I failed to realize that the context was already updated by the time the callback was called and the try_number was 1 step ahead. After fixing the reference I kept it mostly as a fail-safe, but it doesn't seem like it's necessary.

dags/stellar_etl_airflow/utils.py
edualvess: It makes much sense like you suggested. I committed a change for this matter just now.

dags/stellar_etl_airflow/utils.py
sydneynotthecity: It's fine to keep as a fail safe, I was just curious if Airflow was flaking with its logging

PR #421, opened 2024-07-08
Merge Delete from BQ and Insert from GCS to BQ into a Single Atomic Operation for achieving Idempotency

harsha-stellar-data:
<details>

  - Created a Custom Python Task (`build_del_ins_from_gcs_to_bq_task`) to combine the Delete from BQ and Insert from GCS to BQ Tasks.
  - Updated the History Table Export and State table Export DAGs to handle variable assignments, parameter exchange, task lineup updates after the changes.
  - Updated the Airflow variables docs to add `task_sla` and `task_timeout` for the new task.
  - Updated the image of history_table_export and state table export DAGs to show the new task flow.
  - Updated the documentation page to refer to the new task.

  ### PR Structure

        otherwise).
  - [X] This PR's title starts with the JIRA ticket associated with the PR.

  ### Thoroughness

  - [X] I've updated the README with the added features, breaking changes, new instructions on how to use the repository.

</details>

### What

https://stellarorg.atlassian.net/browse/HUBBLE-398

### Why

This PR Changes the way a DELETE and INSERT ops are handled in the history_table_export DAG. Existing method is causing duplicated data issues during restarts and this PR addresses the issue by combining both the operators into one atomic operation thus ensuring a clean up is always performed before an Insert is done to avoid potential duplication.

### Known limitations

[TODO]
enriched_history_operations is still using build_delete_data_task as it is following a different task build_bq_insert_job for the Insert operation. We may have to review these and see if there could be an impact there during the operations being non-atomic


dags/history_tables_dag.py
```diff
@@ -9,30 +9,35 @@
...
     init_sentry,
 )
 
 init_sentry()
 
-
+# Initialize the DAG
 dag = DAG(
-    "history_table_export",
+    "history_tables_export",
```
chowbao, line 36: I don't think we should rename the DAG in. this PR. It requires extra coordination with what is currently running in prod.

This can be done as a followup PR

dags/history_tables_dag.py
```diff
@@ -9,30 +9,35 @@
...
     default_args=get_default_dag_args(),
-    start_date=datetime(2024, 6, 11, 17, 30),
+    start_date=datetime(2024, 7, 8, 16, 00),
     catchup=True,
-    description="This DAG exports trades and operations from the history archive using CaptiveCore. This supports parsing sponsorship and AMMs.",
+    description="This DAG exports trades, asstes, ledgers, operations, transactions, effects from the history archive using CaptiveCore. This supports parsing sponsorship and AMMs.",
```
chowbao, line 40:
```suggestion
    description="This DAG exports trades, assets, ledgers, operations, transactions, effects.",
```

README.md
```diff
@@ -668,6 +669,10 @@ This section contains information about the Airflow setup. It includes our DAG d
 
 [This file](https://github.com/stellar/stellar-etl-airflow/blob/master/dags/stellar_etl_airflow/build_gcs_to_bq_task.py) contains methods for creating tasks that appends information from a Google Cloud Storage file to a BigQuery table. These tasks will create a new table if one does not exist. These tasks are used for history archive data structures, as Stellar wants to keep a complete record of the ledger's entire history.
 
+### **build_del_ins_from_gcs_to_bq_task**
+
+[This file](https://github.com/stellar/stellar-etl-airflow/blob/master/dags/stellar_etl_airflow/build_del_ins_from_gcs_to_bq_task.py) contains methods for deleting data from old partitions if the table exists and also import fresh data from gcs to the corresponding Big Query table. These tasks will create a new table if one does not exist. These tasks are used for history archive data structures, as Stellar wants to keep a complete record of the ledger's entire history.
```
chowbao, line 674:
```suggestion
[This file](https://github.com/stellar/stellar-etl-airflow/blob/master/dags/stellar_etl_airflow/build_del_ins_from_gcs_to_bq_task.py) contains methods for deleting data from a specified BigQuery table according to the batch interval and also imports data from gcs to the corresponding BigQuery table. These tasks will create a new table if one does not exist. These tasks are used for history and state data structures, as Stellar wants to keep a complete record of the ledger's entire history.
```

dags/history_tables_dag.py
```diff
@@ -169,22 +180,210 @@
...
-
-delete_old_op_pub_task = build_delete_data_task(
-    dag, public_project, public_dataset, table_names["operations"], "pub"
-)
-
-
-delete_old_trade_pub_task = build_delete_data_task(
-    dag, public_project, public_dataset, table_names["trades"], "pub"
+del_ins_ops_task = PythonOperator(
+    task_id=f"del_ins_ops_task",
```
chowbao, line 251: Also within the function described in the [previous comment](https://github.com/stellar/stellar-etl-airflow/pull/421/files#r1669125002) you could make it standardize and define a task_id/name instead of needing to manually type out `del_ins_<export task name>_task`. You would still need to pass a name for `<export task name>` or just use the `*_export_task.task_id`

dags/stellar_etl_airflow/build_del_ins_from_gcs_to_bq_task.py
chowbao, line 1: I'm pretty sure we want to to be applied to the state table dag as well.
As you can see it also used the old delete and insert functions [here](https://github.com/stellar/stellar-etl-airflow/blob/master/dags/state_table_dag.py#L98)

harsha-stellar-data:
@chowbao All the suggested changes are taken care. Please review and let me know if anything else needs to be looked at. Thanks

dags/history_tables_dag.py
chowbao: FYI in the future you can just click `Commit suggestion` instead of copy and pasting them locally. It makes a new commit and all you'd have to do is do a `git pull` on your branch locally

dags/history_tables_dag.py
```diff
@@ -47,10 +57,14 @@
         "subtract_data_interval": macros.subtract_data_interval,
         "batch_run_date_as_datetime_string": macros.batch_run_date_as_datetime_string,
     },
-    # sla_miss_callback=alert_sla_miss,
+    sla_miss_callback=alert_sla_miss,
```
chowbao, line 60: This should be left commented out. The sla alerts still don't work as intended and are too noisy right now

dags/history_tables_dag.py
```diff
@@ -59,37 +73,7 @@
 use_captive_core = literal_eval(Variable.get("use_captive_core"))
 txmeta_datastore_path = "{{ var.value.txmeta_datastore_path }}"
 
-
```
chowbao, line 62: What was the purpose of deleting and moving existing tasks around?

For example `time_task = build_time_task(dag, use_testnet=use_testnet, use_futurenet=use_futurenet)` was moved to line 188 and replaced the comment with just `# Define time task
`

dags/history_tables_dag.py
harsha-stellar-data: Will comment it.

dags/history_tables_dag.py
```diff
@@ -47,10 +57,14 @@
         "subtract_data_interval": macros.subtract_data_interval,
         "batch_run_date_as_datetime_string": macros.batch_run_date_as_datetime_string,
     },
-    # sla_miss_callback=alert_sla_miss,
+    sla_miss_callback=alert_sla_miss,
 )
 
+# Initialize batch metadata variables
+batch_id = macros.get_batch_id()
+batch_date = "{{ batch_run_date_as_datetime_string(dag, data_interval_start) }}"
```
chowbao, line 65: This could be moved to the function in https://github.com/stellar/stellar-etl-airflow/pull/421/files#diff-62ce6084a9fcd6873489fbc5700bff8413d481cb580f0fdd2d641b744c68b909

Not sure if macros will create a circular dependency. You might need to only import the macros.<functions> that are needed in each file

dags/history_tables_dag.py
harsha-stellar-data: I will update the proper comments back. Took care of some sections but this and export task are missed when I merged the changes back from my test script to original.

dags/history_tables_dag.py
harsha-stellar-data: batch_id and batch_date are DAG level. So, i thought it would be best to leave them as oarams coming from a DAG rather than a reusable piece of code.

dags/stellar_etl_airflow/build_del_ins_operator.py
```diff
@@ -0,0 +1,84 @@
...
+        "source_object_suffix": source_object_suffix,
+        "partition": True,
+        "cluster": True,
+        "batch_id": batch_id,
+        "batch_date": batch_date,
+        "source_objects": source_objects,
+    }
+
+
+def create_del_ins_task(dag, task_id, task_vars, del_ins_callable):
```
chowbao, line 54: nit: `create_del_ins_task` could probably be merged into `initialize_task_vars` and renamed.

Like right now all `create_del_ins_task` is doing is passing through variables. So in `initialize_task_vars` why return a dict of variables when instead you could pass the variables directly to the PythonOperator?

dags/history_tables_dag.py
```diff
@@ -168,22 +158,43 @@
...
+        data_type,
+        export_task.task_id,
+        batch_id,
+        batch_date,
+        table_names,
+        public_project,
+        public_dataset,
+    )
+    del_ins_tasks[data_type] = create_del_ins_task(
+        dag, f"del_ins_{data_type}_task", task_vars, build_del_ins_from_gcs_to_bq_task
```
chowbao, line 184: The fstring for the task name can be done within `initialize_task_vars`

dags/history_tables_dag.py
chowbao: That's probably fine.

My nit comes from the perspective that `batch_id` and `batch_date` are only used in the new functions for the delete/insert task and the delete/insert task will always need `batch_id` and `batch_date` anyways so why bother passing them as a parameter

dags/stellar_etl_airflow/build_del_ins_operator.py
harsha-stellar-data: Implemented them as separate functions as we can use either of them separately if needed as well. For initialising in any other flows too if required. 

where as the Del Ins task serves the specific purpose. So, separating them seems logical.

chowbao: I'm not sure I agree but I think that's just an opinion.

If that's the case why even have `create_del_ins_task`? All it's doing is returning `PythonOperator`. You could just call `PythonOperator` with the same params

dags/stellar_etl_airflow/build_del_ins_operator.py
```diff
@@ -0,0 +1,84 @@
...
+        task_vars (dict): Task variables.
+        del_ins_callable (callable): The callable function to be used in the PythonOperator.
+
+    Returns:
+        PythonOperator: The created PythonOperator.
+    """
+    return PythonOperator(
+        task_id=task_id,
+        python_callable=del_ins_callable,
+        op_args=[
```
chowbao, line 70: nit: this could have been `op_kwargs` instead of `op_args`

```
op_kwargs=task_vars,
```

dags/stellar_etl_airflow/build_del_ins_operator.py
harsha-stellar-data: sure. We can discuss in the office hours but the whole point of separating these out into a separate script is to provide a level of  abstraction is my understanding. Combining them may be an okay idea for now as the code base is small but long term, i think we should keep them separated.

dags/stellar_etl_airflow/build_del_ins_operator.py
```diff
@@ -0,0 +1,84 @@
...
+    public_project,
+    public_dataset,
+    source_object_suffix="",
+    source_objects=None,
+):
+    """
+    Initialize task variables for data export and import.
+
+    Args:
+        data_type (str): Type of data (e.g., operations, trades).
```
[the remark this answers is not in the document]
harsha-stellar-data: I did not get this suggestion. (eg.,) is the proper form, right?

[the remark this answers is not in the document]
chowbao: Oh it is TIL
https://www.bachelorprint.com/commas/comma-before-or-after-e-g/#:~:text=When%20to%20place%20a%20comma,main%20part%20of%20your%20sentence.

dags/stellar_etl_airflow/build_del_ins_operator.py
harsha-stellar-data: Will update to kwargs. I usyally prefer the params explicitly mentioned but in this case, since the task vars initialization is right above, use of kwargs should be clear.

chowbao (dismissed):

documentation/images/history_table_export.png
chowbao, line 1: Refresh dag image: I think it's stale

documentation/images/state_table_export.png
chowbao, line 1: Refresh dag image: I think it's stale

dags/state_table_dag.py
```diff
@@ -90,240 +102,76 @@
...
+    "trustlines": "/*-trustlines.txt",
+    "contract_data": "/*-contract_data.txt",
+    "contract_code": "/*-contract_code.txt",
+    "config_settings": "/*-config_settings.txt",
+    "ttl": "/*-ttl.txt",
+}
+
+# filter for only the required tables pertaining to the DAG
+table_id_and_suffixes = {
+    key: suffix
```
chowbao, line 131: nit: I don't think this needs to create a new dict. `source_object_suffix_mapping` and `table_id_and_suffixes` are effectively the same thing I think.

Also this kind of fails silently if the `key` is not in `table_names` whereas I think right now if `"accounts"` wasn't defined in the airflow var for `table_names` it would give us an error

dags/stellar_etl_airflow/build_del_ins_from_gcs_to_bq_task.py
```diff
@@ -0,0 +1,204 @@
...
+        )
+        time_partition["type"] = partition_fields["type"]
+        time_partition["field"] = partition_fields["field"]
+
+    staging_table_suffix = ""
+    if table_name == "history_assets":
+        staging_table_suffix = "_staging"
+
+    if table_id in history_tables:
+        schema_fields = read_local_schema(f"history_{table_id}")
```
chowbao, line 140: nit: I think this can be updated to use `table_names`
Also funnily enough, setting `schema_fields` can now be done outside the if/else logic

```suggestion
        schema_fields = read_local_schema(f"{table_names}")
```

harsha-stellar-data: Agreed. Moved it outside the conditional statements and defined it once outside.

documentation/images/history_table_export.png
harsha-stellar-data: Will take care of this

documentation/images/state_table_export.png
harsha-stellar-data: Will take care of this

dags/state_table_dag.py
harsha-stellar-data: I will remove this and add a check if the keys are missing in the table_names variable.

chowbao (approved):

PR #433, opened 2024-07-16
Update airflow resources and split out dbt tests and alerts

chowbao:
<details>

### PR Structure

      otherwise).

### Thoroughness

</details>

### What

Reducing airflow resources to reduce costs and splitting out tests from the elementary alert DAG

### Why

* Airflow resources were originally set for using captive core. Now that we have moved to CDP and observed the actual resources needed. We can greatly reduce the resources requested
* The elementary alert DAG has a dependency on a dbt test. If this test fails the elementary alerts won't actually fire. Separating them out should avoid this

### Known limitations

N/A

### DIFF
--- a/airflow_variables_dev.json
+++ b/airflow_variables_dev.json
@@ -284,21 +284,21 @@
   "public_project": "test-hubble-319619",
   "resources": {
     "dbt": {
-      "requests": {
+      "requests": {
         "cpu": "0.2",
         "ephemeral-storage": "500Mi",
         "memory": "250Mi"
       }
     },
     "default": {
-      "requests": {
+      "requests": {
         "cpu": "0.2",
         "ephemeral-storage": "500Mi",
         "memory": "250Mi"
       }
     },
     "stellar-etl": {
-      "requests": {
+      "requests": {
         "cpu": "0.2",
         "ephemeral-storage": "500Mi",
         "memory": "250Mi"
--- a/airflow_variables_dev.json
+++ b/airflow_variables_dev.json
@@ -283,21 +283,21 @@
   "public_dataset": "test_crypto_stellar",
   "public_project": "test-hubble-319619",
   "resources": {
-    "default": {
+    "dbt": {
       "requests": {
         "cpu": "0.2",
         "ephemeral-storage": "500Mi",
         "memory": "250Mi"
       }
     },
-    "stellar-etl": {
+    "default": {
       "requests": {
         "cpu": "0.2",
         "ephemeral-storage": "500Mi",
         "memory": "250Mi"
       }
     },
-    "dbt": {
+    "stellar-etl": {
       "requests": {
         "cpu": "0.2",
         "ephemeral-storage": "500Mi",
--- a/airflow_variables_dev.json
+++ b/airflow_variables_dev.json
@@ -285,24 +285,24 @@
   "resources": {
     "dbt": {
       "requests": {
-        "cpu": "0.2",
-        "ephemeral-storage": "500Mi",
-        "memory": "250Mi"
-      }
+        "cpu": "0.2",
+        "ephemeral-storage": "500Mi",
+        "memory": "250Mi"
+      }
     },
     "default": {
-      "requests": {
-        "cpu": "0.2",
-        "ephemeral-storage": "500Mi",
-        "memory": "250Mi"
-      }
+      "requests": {
+        "cpu": "0.2",
+        "ephemeral-storage": "500Mi",
+        "memory": "250Mi"
+      }
     },
     "stellar-etl": {
-      "requests": {
-        "cpu": "0.2",
-        "ephemeral-storage": "500Mi",
-        "memory": "250Mi"
-      }
+      "requests": {
+        "cpu": "0.2",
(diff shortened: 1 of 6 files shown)


airflow_variables_dev.json
```diff
@@ -283,33 +283,26 @@
...
-        "memory": "900Mi"
-      }
+      "requests": {
+        "cpu": "0.2",
+        "ephemeral-storage": "500Mi",
+        "memory": "250Mi"
+      }
     },
-    "state": {
+    "stellar-etl": {
```
chowbao, line 293: Added new `stellar-etl`and `dbt` resources. This will make it easy to adjust resources without adjusting resources for every dag/task

dags/dbt_singular_tests_dag.py
chowbao, line 1: Technically this could have been added to the existing dbt eho/state DAG. I opted to separate it out because it is purely for running anomaly tests/other tests which felt different enough to warrant a standalone DAG

dags/history_tables_dag.py
```diff
@@ -112,6 +112,7 @@
     use_gcs=True,
     use_captive_core=use_captive_core,
     txmeta_datastore_path=txmeta_datastore_path,
+    resource_cfg="stellaretl",
```
amishas157, line 115: Where is the config defined? I see build_export_task imports it from `f"{{{{ var.json.resources.{resource_cfg}.requests | container_resources }}}}"`

dags/dbt_data_quality_alerts_dag.py
```diff
@@ -28,14 +28,7 @@
...
-        command_type="test",
-        tag="singular_test",
-    )
-    singular_tests_elementary_alerts = elementary_task(dag, "dbt_data_quality")
-    start_tests = EmptyOperator(task_id="start_tests_task")
+    # Trigger elementary
+    elementary_alerts = elementary_task(dag, "dbt_data_quality")
 
-    # DAG task graph
-    start_tests >> singular_tests >> singular_tests_elementary_alerts
```
amishas157, line 41: This looks good. But curious why were we using empty operator?

Also, different DAG is okay, but we can create independent tasks in same DAG as well, correct?

```
elementary_alerts
singular_tests_elementary_alerts
```

dags/history_tables_dag.py
amishas157: oh ignore my q. Looks it is defined in airflow_variables_dev/prod.json

airflow_variables_prod.json
```diff
@@ -281,32 +281,25 @@
...
+        "memory": "1Gi"
       }
     },
     "default": {
       "requests": {
-        "cpu": "3.5",
+        "cpu": "0.5",
         "ephemeral-storage": "1Gi",
-        "memory": "5Gi"
+        "memory": "1Gi"
```
amishas157, line 295: For my own knowledge, are the historical resource usage metrics tracked somewhere?

amishas157 (approved):
Looks good to me. Left some q inline for my understanding

dags/dbt_data_quality_alerts_dag.py
```diff
@@ -18,7 +18,7 @@
     default_args=get_default_dag_args(),
     start_date=datetime(2024, 6, 25, 0, 0),
     description="This DAG runs dbt tests and Elementary alerts at a half-hourly cadence",
-    schedule="*/15,*/45 * * * *",  # Runs every 15th minute and every 45th minute
+    schedule="15,45 * * * *",  # Runs every 15th minute and every 45th minute
```
sydneynotthecity, line 21: Is there a reason we're running at the 15th and 45th minutes instead of `*/30 ****`?

airflow_variables_prod.json
sydneynotthecity, line 1: If we're no longer separating resources out between dbt, default and stellaretl, is it worth us storing configs beyond the `default`?

airflow_variables_prod.json
chowbao: I don't believe it is. In theory you could find it in GCP logs

dags/dbt_data_quality_alerts_dag.py
chowbao: Yes. So the other dbt DAGs run on `*/30` with task run times around 10 mins. If we were to run the quality alerts at the same time we would have a slightly larger gap in time before we are notified. Running at a 15 min offset to those runs _hopefully_ alerts faster

airflow_variables_prod.json
chowbao: I think it's worth it. I was going to delete all the resources and just make everything run with the default settings but I have a theory that dbt and stellar-etl both have different resource requirements. So in the case we do need to bump up/down resources for one service we wouldn't necessarily OOM or over-allocate the other service

dags/dbt_data_quality_alerts_dag.py
chowbao: Alternatively I don't see the harm in running every 15 mins like it has been lol

airflow_variables_prod.json
sydneynotthecity: 👌 sounds good to me. Your theory sounds plausible

PR #489, opened 2024-09-06
Add 30 day filter to audit_log

chowbao:
<details>

### PR Structure

      otherwise).

### Thoroughness

</details>

### What

Adding a 30 day filter to the `audit_log` query. Was originally failing because it is a full table truncate_write where full history  was causing the following error:

```
google.api_core.exceptions.BadRequest: 400 Resources exceeded during query execution: The query could not be executed in the allotted memory. Peak usage: 147% of limit.
Top memory consumer(s):
  aggregate functions and GROUP BY clauses: 99%
  other/unattributed: 1%
```

### Why

To fix running the audit_log_dag

### Known limitations

The audit_log table will no longer contain full history but instead only 30 days of information. The source tables that produce the audit_log however will continue to have full history. The query will need to be manually run with the correct timestamp window if information more than 30 days ago is needed

### DIFF
--- a/dags/queries/audit_log.sql
+++ b/dags/queries/audit_log.sql
@@ -229,6 +229,8 @@ WITH jobChangeEvent AS (
     JSON_EXTRACT_SCALAR(protopayload_auditlog.metadataJson, '$.jobChange.after') AS jobChangeAfter,
     REGEXP_EXTRACT(protopayload_auditlog.metadataJson, r'BigQueryAuditMetadata","(.*?)":') AS eventName,
   FROM {project_id}.{dataset_id}.cloudaudit_googleapis_com_data_access
+  where true
+    and timestamp >= current_timestamp - interval 30 day
 ),
 /*
  * TableCreation: Table creation event.
@@ -272,6 +274,8 @@ tableCreationEvent AS (
     JSON_EXTRACT_SCALAR(protopayload_auditlog.metadataJson,
       '$.tableCreation.reason')  AS tableCreationReason,
   FROM {project_id}.{dataset_id}.cloudaudit_googleapis_com_activity
+  where true
+    and timestamp >= current_timestamp - interval 30 day
 ),
 /*
  * TableChange: Table metadata change event
@@ -317,6 +321,8 @@ tableChangeEvent AS (
     CAST(JSON_EXTRACT_SCALAR(protopayload_auditlog.metadataJson,
       '$.tableChange.truncated') AS BOOL) AS tableChangeTruncated
   FROM {project_id}.{dataset_id}.cloudaudit_googleapis_com_activity
+  where true
+    and timestamp >= current_timestamp - interval 30 day
 ),
 /*
  * TableDeletion: Table deletion event
@@ -337,6 +343,8 @@ tableDeletionEvent AS (
     JSON_EXTRACT_SCALAR(protopayload_auditlog.metadataJson,
       '$.tableDeletion.reason') AS tableDeletionReason,
   FROM {project_id}.{dataset_id}.cloudaudit_googleapis_com_activity
+  where true
+    and timestamp >= current_timestamp - interval 30 day
 ),
 /*
  * TableDataRead: Data from tableDataRead audit logs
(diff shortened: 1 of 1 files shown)


dags/queries/audit_log.sql
```diff
@@ -229,6 +229,8 @@ WITH jobChangeEvent AS (
     JSON_EXTRACT_SCALAR(protopayload_auditlog.metadataJson, '$.jobChange.after') AS jobChangeAfter,
     REGEXP_EXTRACT(protopayload_auditlog.metadataJson, r'BigQueryAuditMetadata","(.*?)":') AS eventName,
   FROM {project_id}.{dataset_id}.cloudaudit_googleapis_com_data_access
+  where true
+    and timestamp >= current_timestamp - interval 30 day
```
chowbao, line 233: `timestamp` is the name of the column these tables are partitioned on

sydneynotthecity:
> The audit_log table will no longer contain full history but instead only 30 days of information. The source tables that produce the audit_log however will continue to have full history. The query will need to be manually run with the correct timestamp window if information more than 30 days ago is needed

I have an alternate proposal: the `build_bq_insert_job` operator also supports write disposition of `WRITE_APPEND`. What do you think about tweaking the date filter to only load the last day of audit log history so that we can maintain full history within the tables? 

The date logic will need to be adjusted somewhere to either pass in airflow date (making the pipeline idempotent), or at minimum, ensure that only 1 day of history is fetched at a time, ie:
```
where timestamp >= timestamp_sub(timestamp_trunc(current_timestamp(), day))
  and timestamp < timestamp_trunccurrent_timestamp(), day)
```

chowbao:
> > The audit_log table will no longer contain full history but instead only 30 days of information. The source tables that produce the audit_log however will continue to have full history. The query will need to be manually run with the correct timestamp window if information more than 30 days ago is needed
> 
> I have an alternate proposal: the `build_bq_insert_job` operator also supports write disposition of `WRITE_APPEND`. What do you think about tweaking the date filter to only load the last day of audit log history so that we can maintain full history within the tables?
> 
> The date logic will need to be adjusted somewhere to either pass in airflow date (making the pipeline idempotent), or at minimum, ensure that only 1 day of history is fetched at a time, ie:
> 
> ```
> where timestamp >= timestamp_sub(timestamp_trunc(current_timestamp(), day), interval 1 day)
>   and timestamp < timestamp_trunccurrent_timestamp(), day)
> ```

Yeah I can adjust it. What do you think of using the following logic instead? 
I'm not sure I see the benefit of running it idempotently. Especially because this runs with `catchup=False` meaning the above logic could potentially miss data if a day is skipped or the airflow execution date for a date could be missing (dag runs daily).

```
timestamp >= (select max(timestamp) from `project.dataset.audit_log`)
```

sydneynotthecity:
Yeah I think that's good enough for this DAG. We can always manually intervene if needed for hairier failures. Since the upstream source maintains history, we have ways to backfill and recover as needed.

Other than the query recently blowing up, this DAG has been very stable so we don't need to be as precise with recovery.

nit: I think the logic needs to be non exclusive of the endpoint: `timestamp > (select max(timestamp) from `project.dataset.audit_log`)` otherwise you'll be duplicating the data that exactly matches the date bound

dags/queries/audit_log.sql
sreuland: what is significance for AND'ing `true` literal in the clause, can be removed?

chowbao: It's a stylistic thing when writing queries just like using `,` before or after a column
For example let's say you have a bunch of filters you want to apply to the query. All of these will behave exactly the same way (query planner will optimize it to the same query plan)
```
where filter1 >= 1
  and filter2 >= 2
  and filter3 >= 3
```

```
where true
  and filter1 >= 1
  and filter2 >= 2
  and filter3 >= 3
```

```
where filter1 >= 1 and
  filter2 >= 2 and
  filter3 >= 3
```

```
where true and
  filter1 >= 1 and
  filter2 >= 2 and
  filter3 >= 3
```

I just personally find this style to be my preference because when you are debugging or exploring a table you can easily add new filters or comment out filters without need to add addition logic
```
where true
  --and filter1 >= 1
  and filter2 >= 2
  and filter3 >= 3
```
vs
```
where
 -- filter1 >= 1
  filter2 >= 2 -- needed to delete the and as well
  and filter3 >= 3
```

sydneynotthecity (approved):
RIP, full history

After experimenting with `WRITE_APPEND`, we realized the query actually runs into quotas and still will not execute properly in an incremental mode. We opted to stay with `WRITE_TRUNCATE` option originally proposed.

PR #506, opened 2024-10-03
Add dag to create elementary report

amishas157:
<details>

### PR Structure

      otherwise).

### Thoroughness

</details>

### What

This PR adds a new DAG to generate elementary report and send it to slack. The DAG will be run once a week (on Monday) and generate report for last 7 days.

### Why

The obervability report will help us to monitor any pending test failures during the week.

### Known limitations

The edr command is lil wonky with memory requirements. It is a known issue and observed by many community members as well.
https://github.com/elementary-data/elementary/issues/761 
Even though above is closed, but it is not fully resolved. Also, it requires high ephemeral storage since it downloads all the data and then compute the report.

### Testing:

- ✅ It takes 7-8 min to finish generating the report
- ✅ Screenshot 2024-11-12 at 1 35 19 PM
- ❓ High memory usage check from gcloud
Screenshot 2024-11-12 at 1 27 50 PM


dags/elementary_report_dag.py
```diff
@@ -0,0 +1,38 @@
...
+)
+
+init_sentry()
+
+with DAG(
+    "elementary_report",
+    default_args=get_default_dag_args(),
+    start_date=datetime(2024, 6, 25, 0, 0),
+    description="This DAG creates elementary report and send it to slack",
+    schedule="*/60 * * * *",  # Runs every 60 minutes
```
sydneynotthecity, line 21: I think hourly might be too frequent for this to send reports to slack. I would expect either a daily or weekly summary report. Is there a mockup where we can see what this report looks like?

sydneynotthecity (commented):
I'm curious about the implications of running this DAG hourly. Will the report only contain data for the past day? Is it only sent if there are failures?

dags/elementary_report_dag.py
amishas157: Yeah, it is still WIP. I am hitting OOMkilled error and need to do more testing.

airflow_variables_dev.json
```diff
@@ -124,7 +124,7 @@
     "partnership_assets__account_holders_activity_fact": false,
     "partnership_assets__asset_activity_fact": false
   },
-  "dbt_image_name": "stellar/stellar-dbt:53375b5f9",
+  "dbt_image_name": "stellar/stellar-dbt-dev:dba580d7c",
```
amishas157, line 127: TODO: Update this image once final dbt image is created. Also update corresponding var in prod

airflow_variables_dev.json
```diff
@@ -288,21 +288,28 @@
     "dbt": {
       "requests": {
         "cpu": "1",
-        "ephemeral-storage": "500Mi",
+        "ephemeral-storage": "1Gi",
```
amishas157, line 291: The default value of `ephemeral-storage` is 1 Gi. We were not propagating this value
https://github.com/stellar/stellar-etl-airflow/pull/506/files#diff-f7be361360b2906f863ccfd863e335513b7b5ccc42728f16c2ae1ef43a798465R36

sydneynotthecity (approved):
one suggestion but looks good. We won't need to upgrade worker size, correct?

airflow_variables_dev.json
amishas157: Not really, good to 🔪

amishas157:
> one suggestion but looks good. We won't need to upgrade worker size, correct?

No, we don't need to

airflow_variables_dev.json
amishas157: Done

PR #507, opened 2024-10-07
Generate avro files from bq tables

chowbao:
<details>

### PR Structure

      otherwise).

### Thoroughness

</details>

### What

Adding DAG to generate hourly AVRO files from BQ tables

### Why

Needed a way to generate frontfill AVRO files for external analytics partnerships

### Known limitations

* Opted for individual table.sql files instead of programmatically generated through airflow/python even though this duplicates SQL. Main reason being it is unknown what kind of query logic that will need to be propagated to the queries to generate the AVRO files. Example is history_effects needing to 1. exclude columns 2. flatten a column 3. exclude a column from the flattened columns. It seems easier to manage at the individual table SQL level than adding function parameters in airflow
* Did not include a monthly time window backfill DAG
* `history_effects` and `history_operations` currently not run until confirmation that the tables works for external partners

### DIFF
--- a/dags/generate_avro_files_dag.py
+++ b/dags/generate_avro_files_dag.py
@@ -1,11 +1,12 @@
 from datetime import datetime
 
 from airflow import DAG
-from kubernetes.client import models as k8s
+from stellar_etl_airflow import macros
+from stellar_etl_airflow.build_cross_dependency_task import build_cross_deps
 from stellar_etl_airflow.build_bq_generate_avro_job_task import (
     build_bq_generate_avro_job,
 )
-from stellar_etl_airflow.build_cross_dependency_task import build_cross_deps
+from airflow.operators.dummy import DummyOperator
 from stellar_etl_airflow.default import (
     alert_sla_miss,
     get_default_dag_args,
@@ -15,16 +16,18 @@ from stellar_etl_airflow.default import (
 init_sentry()
 
 dag = DAG(
-    "generate_avro_files",
+    "generate_avro",
     default_args=get_default_dag_args(),
-    start_date=datetime(2024, 10, 1, 0, 0),
-    description="This DAG generates AVRO files from BQ tables",
-    schedule_interval="0 * * * *",  # Runs every hour
-    user_defined_filters={
-        "container_resources": lambda s: k8s.V1ResourceRequirements(requests=s),
-    },
-    max_active_runs=5,
+    start_date=datetime(2024, 10, 1, 1, 0),
     catchup=True,
+    description="This DAG generates AVRO files from BQ tables",
+    schedule_interval="0 * * * *",
+    render_template_as_native_obj=True,
+    user_defined_macros={
+        "subtract_data_interval": macros.subtract_data_interval,
+        "batch_run_date_as_datetime_string": macros.batch_run_date_as_datetime_string,
+        "batch_run_date_as_directory_string": macros.batch_run_date_as_directory_string,
+    },
     sla_miss_callback=alert_sla_miss,
 )
 
@@ -32,11 +35,14 @@ public_project = "{{ var.value.public_project }}"
 public_dataset = "{{ var.value.public_dataset }}"
 gcs_bucket = "{{ var.value.avro_gcs_bucket }}"
 
+
 # Wait on ingestion DAGs
 wait_on_history_table = build_cross_deps(
     dag, "wait_on_ledgers_txs", "history_table_export"
 )
 wait_on_state_table = build_cross_deps(dag, "wait_on_state_table", "state_table_export")
+
+dummy_task = DummyOperator(task_id='dummy_task', dag=dag)
 
 # Generate AVRO files
 avro_tables = [
@@ -55,7 +61,7 @@ avro_tables = [
 ]
 
 for table in avro_tables:
-    task = build_bq_generate_avro_job(
+    avro_task = build_bq_generate_avro_job(
         dag=dag,
         project=public_project,
         dataset=public_dataset,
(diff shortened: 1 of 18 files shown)


dags/generate_avro_files_dag.py
```diff
@@ -0,0 +1,63 @@
...
+    get_default_dag_args,
+    init_sentry,
+)
+
+init_sentry()
+
+dag = DAG(
+    "generate_avro_files",
+    default_args=get_default_dag_args(),
+    start_date=datetime(2024, 10, 1, 0, 0),
```
chowbao, line 18: AVRO files for < 10/1/24 are already generated

sydneynotthecity: Do you think it's easier to manually backfill Oct 1 - Oct 7 data, then frontfill hourly from there? Any concerns about catching the last 7 days up?

dags/queries/generate_avro/history_operations.sql
```diff
@@ -0,0 +1,19 @@
...
+    format = 'avro',
+    overwrite=true
+    )
+as (
+  select
+    *
+    except(details, details_json, batch_id, batch_insert_ts, batch_run_date),
+    details.*
+    except(claimants, type),
+    details.type as details_type
```
sydneynotthecity, line 13: is this soroban_operation_type? Do you think we need to rename this column to be clearer? Might be worth discussing with Andre

dags/queries/generate_avro/history_trades.sql
```diff
@@ -0,0 +1,17 @@
...
+    )
+as (
+  select
+    *
+    except(ledger_closed_at, batch_id, batch_insert_ts, batch_run_date),
+    ledger_closed_at as closed_at
+  from {project_id}.{dataset_id}.history_trades
+  where true
+    and closed_at >= '{batch_run_date}'
+    and closed_at < '{next_batch_run_date}'
```
sydneynotthecity, line 15: `closed_at` is not on the table, i don't think this will run

dags/stellar_etl_airflow/build_bq_generate_avro_job_task.py
```diff
@@ -0,0 +1,66 @@
...
+    )
+    uri_datetime = (
+        "{{ batch_run_date_as_directory_string(dag, data_interval_start) }}"
+    )
+    uri = f"gs://{gcs_bucket}/avro/{table}/{uri_datetime}/*.avro"
+    sql_params = {
+        "project_id": project,
+        "dataset_id": dataset,
+        "batch_run_date": batch_run_date,
+        "prev_batch_run_date": prev_batch_run_date,
```
sydneynotthecity, line 40: Is this used in the SQL query? I didn't see the parameter anywhere in the SQLs

sydneynotthecity (commented):
Nit: I noticed the SQL files are not filtering based on partition key. These tables aren't partitioned by `closed_at` _yet_. Can you check the difference in processing bytes/execution time by excluding partition key `batch_run_date`? Might be negligible, but I have concerns on larger tables, like `history_transactions` and `trust_lines`.

Looks good overall, just want to confirm the queries are performant before we run them hourly :sweat-smile:

dags/generate_avro_files_dag.py
chowbao: Hmm we can. I wonder if that would be confusing for Dune though. Like 10/1 wouldn't be the full month but instead there would be 10/1 and then another folder for 10/8, 10/9, 10/10, etc...

> Any concerns about catching the last 7 days up?
Nope

dags/queries/generate_avro/history_operations.sql
chowbao: > is this soroban_operation_type?
Yes it is

> Do you think we need to rename this column to be clearer?
Yeah we can. I'll regenerate the files and mention it to Andre

dags/stellar_etl_airflow/build_bq_generate_avro_job_task.py
chowbao: `prev_batch_run_date` is not. I'll remove it

chowbao:
> Nit: I noticed the SQL files are not filtering based on partition key. These tables aren't partitioned by `closed_at` _yet_. Can you check the difference in processing bytes/execution time by excluding partition key `batch_run_date`? Might be negligible, but I have concerns on larger tables, like `history_transactions` and `trust_lines`.
> 
> Looks good overall, just want to confirm the queries are performant before we run them hourly :sweat-smile:

Ahh I can add `batch_run_date` filter as well

> Can you check the difference in processing bytes/execution time by excluding partition key `batch_run_date`?

opting to not check this because it makes sense to use the partition column

dags/queries/generate_avro/history_trades.sql
chowbao: Updated to `ledger_closed_at`

dags/generate_avro_files_dag.py
```diff
@@ -0,0 +1,74 @@
...
+gcs_bucket = "{{ var.value.avro_gcs_bucket }}"
+
+
+# Wait on ingestion DAGs
+wait_on_history_table = build_cross_deps(
+    dag, "wait_on_ledgers_txs", "history_table_export"
+)
+wait_on_state_table = build_cross_deps(dag, "wait_on_state_table", "state_table_export")
+
+dummy_task = DummyOperator(task_id="dummy_task", dag=dag)
```
chowbao, line 45: Needed to add a `dummy_task` so that Airflow generates the avro_table loop/dependencies correctly 🤮

sydneynotthecity: oh really! that's gross

sydneynotthecity (approved):

PR #517, opened 2024-10-29
Add workflow to calculate schema changes

amishas157:
<details>

### PR Structure

      otherwise).

### Thoroughness

</details>

### What

 This PR creates a github workflow which:
- runs once a day
- queries for changes in marts schemas
- If any diff, create a branch and push the changes 
- Makes a pull request to update changelog file. Sample PR (https://github.com/stellar/stellar-etl-airflow/pull/526)

Since the results from script will be deterministic, no new PRs will be created if no schema has changed

### Why

To track data schema changes.

I will follow up with 2 PRs:
- Add support to list models added/deleted
- Make CI job to do diff for public datasets schema (from /schemas dir) and create changelog/source_data.md

### Question

Does the format for changelog look alright? Happy to discuss other ideas


scripts/update_dbt_marts_schema_changelog.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+result=$(bq query --format=prettyjson --nouse_legacy_sql \
+    'SELECT
+        date(detected_at) as date
+        , table_name
+        , sub_type as operation
+        , ARRAY_AGG(column_name) as columns
+ FROM
+   `hubble-261722`.elementary.alerts_schema_changes
```
chowbao, line 10: nit: Can we parameterize `hubble-261722` to a `{project}` variable?

.github/workflows/update_dbt_marts_schema_changelog.yml
```diff
@@ -0,0 +1,73 @@
+name: Update changelog for DBT marts
+
+on:
+  schedule:
+    - cron: "0 23 * * 1-5" # This will run at 11 PM UTC, which is 5 PM CST
```
chowbao, line 5: Why not run on weekends?

.github/workflows/update_dbt_marts_schema_changelog.yml
```diff
@@ -0,0 +1,73 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_protected == 'true' && github.sha || github.ref }}-{{ github.event_name }}
+  cancel-in-progress: true
+
+env:
+  IS_RECENCY_AIRFLOW_TASK: "false"
```
chowbao, line 12: Is this needed or is it extra?

.github/workflows/update_dbt_marts_schema_changelog.yml
amishas157: Not really any point, who is there to merge. Monday run will catchup anyways

.github/workflows/update_dbt_marts_schema_changelog.yml
amishas157: extra, lol. I started working in dbt repo and moved over here. Gonna 🔪 it

.github/workflows/update_dbt_marts_schema_changelog.yml
```diff
@@ -0,0 +1,73 @@
...
+        if: steps.commit_changes.outputs.changes_committed == 'true'
+        run: |
+          git push origin ${{ steps.create_branch.outputs.branch }}
+
+      - name: Create Pull Request
+        if: steps.commit_changes.outputs.changes_committed == 'true'
+        run: |
+          gh pr create -B master -H ${{ steps.create_branch.outputs.branch }} \
+          --title 'Merge ${{ steps.create_branch.outputs.branch }} into master' \
+          --body 'Created by GitHub action'
```
amishas157, line 71:
```suggestion
          --title 'Update schema Changelog for marts' \
          --body 'This PR updates schema changelog for marts. Also, this is an auto-generated PR'
```

.github/workflows/update_dbt_marts_schema_changelog.yml
```diff
@@ -0,0 +1,73 @@
...
+          git config --local user.name "GitHub Action"
+          BRANCH_NAME="update-data-schema-changelog-${{ github.run_id }}"
+          git checkout -b $BRANCH_NAME
+          echo "::set-output name=branch::$BRANCH_NAME"
+
+      - name: Run Bash Script
+        run: |
+          cd $GITHUB_WORKSPACE
+          output=$(. scripts/update_dbt_marts_schema_changelog.sh)
+          echo "$output" > changelog/dbt_marts.md
```
chowbao, line 47: Hmm is it possible to make this append instead of overwrite the entire changelog each run?

There might be a time we delete the `elementary.alerts_schema_changes` table like the times we have deleted the whole `elementary` dataset because it was broken or elementary was doing something unexpected. 

If it's too hard to make this append instead of overwrite we should remember to keep a copy of `elementary.alerts_schema_changes` otherwise we'd lose changelog history

.github/workflows/update_dbt_marts_schema_changelog.yml
chowbao: Ah fair enough

.github/workflows/update_dbt_marts_schema_changelog.yml
amishas157: hrm this process is pretty dependent on `elementary.alerts_schema_changes`. Even if we were to delete this dataset ever, that in itself will affect changelog computing part. So ideally we should not delete that dataset.

However as far as append is concerned, we could
- parse tracked date if any
- add date filter to get data between tracked date and today
- update the tracked date as max(date)

Structure of .md file:
```
Date: 2024-10-10

Table 1

Date: 2024-09-01

Table 2
```

amishas157: But also, I think it's quite a bit of text formatting. We can leave it as is and can revisit later if needed. In the first place, I don't think we are going to see a lot of schema changes in general.

.github/workflows/update_dbt_marts_schema_changelog.yml
amishas157: Addressed in https://github.com/stellar/stellar-etl-airflow/pull/517/commits/8b420dc1a2c96788290cb4764e4b373e859b22c3

scripts/update_dbt_marts_schema_changelog.sh
amishas157: Addressed in https://github.com/stellar/stellar-etl-airflow/pull/517/commits/8b420dc1a2c96788290cb4764e4b373e859b22c3

.github/workflows/update_dbt_marts_schema_changelog.yml
chowbao: Yup sounds good

chowbao (approved):

PR #539, opened 2024-11-11
Remove unused library imports, update dbt task

sydneynotthecity:
<details>

### PR Structure

      otherwise).

### Thoroughness

</details>

### What

Updating orchestration flows for the dbt marts. Historically, all staging tables and intermediate tables had to be manually tagged in order to execute for a domain pipeline (ie, fee_stats, enriched_history). This required the developer to remember to tag all models appropriately, which led to mistagged and forgotten models. The flow now uses `+` operator where appropriate to orchestrate pipelines.

I also cleaned up library imports. 

Also disabled the scd_snapshot state and soroban workflows as the tables are broken and should not be used.

### Why

Staging and intermediate models do not need tagging if node selection is inclusive of upstream models.

### Known limitations

[TODO or N/A]


sydneynotthecity:
TODO: update dbt images once built for testing

dags/stellar_etl_airflow/build_dbt_task.py
```diff
@@ -63,7 +63,7 @@ def dbt_task(
     model_name=None,
     tag=None,
     flag="select",
-    operator="",
+    operator="+",
```
amishas157, line 66: This function is also used in tasks such as recency test and singular test, where we may not need the operator.
My recommendation would be to not make `+` as default operator but rather pass it from the caller end.

https://github.com/stellar/stellar-etl-airflow/blob/b29125b00d8ec1d48573c8c10999614d015fee0b/dags/dbt_recency_tests_dag.py#L32-L37

sydneynotthecity: That's fair. Will revert the default and pass in DAGs

amishas157 (approved):

airflow_variables_dev.json
```diff
@@ -124,7 +124,7 @@
     "partnership_assets__account_holders_activity_fact": false,
     "partnership_assets__asset_activity_fact": false
   },
-  "dbt_image_name": "stellar/stellar-dbt:96cd862b1",
+  "dbt_image_name": "stellar/stellar-dbt-dev:4b8a2ecc4",
```
sydneynotthecity, line 127: Will swap once finished testing in test

dags/dbt_enriched_base_tables_dag.py
```diff
@@ -35,9 +35,9 @@
 
 # DBT models to run
 enriched_history_operations_task = dbt_task(
-    dag, tag="enriched_history_operations", excluded="singular_test"
```
sydneynotthecity, line 38: singular test is set through a env var

dags/dbt_stellar_marts_dag.py
```diff
@@ -33,20 +33,39 @@
...
+liquidity_pools_values_task = dbt_task(
+    dag, tag="liquidity_pools_value", operator="+", excluded="stellar_dbt_public"
+)
+liquidity_pools_value_history_task = dbt_task(
+    dag,
+    tag="liquidity_pools_value_history",
+    operator="+",
+    excluded="stellar_dbt_public",
+)
+trade_agg_task = dbt_task(dag, tag="trade_agg", operator="+")
```
sydneynotthecity, line 53: trade agg is defined in the public project

dags/dbt_stellar_marts_dag.py
```diff
@@ -33,20 +33,39 @@
...
 fee_stats_agg_task = dbt_task(dag, tag="fee_stats")
```
amishas157, line 54: We do not exclude stellar_dbt_public when mart is part of public repo.
When do we **not** want to put `operator='+'`?

amishas157 (commented):
One [question on the usage of operator.](https://github.com/stellar/stellar-etl-airflow/pull/539#discussion_r1846943720) Otherwise looks good to me

dags/dbt_stellar_marts_dag.py
sydneynotthecity: imo, only when the mart uses _only_ tables built in the `enriched_history` or `current_state` data pipelines. ie, if your model only uses `enriched_history_operations` as a source, has no intermediate tables, etc. This is why I left default operator settings for `fee_stats_agg` and `network_stats_agg` 

The models cannot have custom intermediate or staging models for this to work.

amishas157: Got it. So, we check lineage. If mart is not dependent on intermediate tables / custom sources, we use default operator to avoid race conditions. Otherwise, we use `+` operator.
That should work. But also, let's continue our discussion in ofc hours for using other ways of model selection.

sydneynotthecity: i agree. This is still imperfect and I think we can come up with an better alternative

PR #550, opened 2024-12-06
Add DAG for exporting retool data

amishas157:
<details>

### PR Structure

      otherwise).

### Thoroughness

</details>

### What

This PR adds DAG to fetch custom data from retool API. It triggers command in stellar-internal , pull data created/updated within given time range.
Existing batch_id is deleted and then re-created(This is to keep idempotent results while retries)

### Why

To support new data ingestion

### Known limitations

Not a limitation, but open to renaming the dag if it too generic. Probable options:
- external_data_dag_retool
- export_retool
- external_data_dag_daily (This can be used to import any external data which runs at daily cadence)

### Testing

Tested the airflow in test-hubble and records look loaded in BQ table

### DIFF
--- a/dags/external_data_dag.py
+++ b/dags/external_data_dag.py
@@ -22,6 +22,7 @@ from stellar_etl_airflow.build_internal_export_task import (
     get_airflow_metadata,
 )
 from stellar_etl_airflow.default import get_default_dag_args, init_sentry
+from stellar_etl_airflow.utils import access_secret
 
 init_sentry()
 
@@ -65,7 +66,9 @@ retool_export_task = build_export_task(
         "{{ subtract_data_interval(dag, data_interval_end).isoformat() }}",
     ],
     use_gcs=True,
-    env_vars={"RETOOL_API_KEY": "{{ var.value.retool_api_key }}"},
+    env_vars={
+        "RETOOL_API_KEY": access_secret("retool-api-key", "default"),
+    },
 )
 
 
--- a/dags/external_data_dag.py
+++ b/dags/external_data_dag.py
@@ -40,6 +40,9 @@ dag = DAG(
         "alias": "external",
     },
     render_template_as_native_obj=True,
+    user_defined_macros={
+        "batch_run_date_as_datetime_string": macros.batch_run_date_as_datetime_string,
+    },
     user_defined_filters={
         "fromjson": lambda s: loads(s),
         "container_resources": lambda s: k8s.V1ResourceRequirements(requests=s),
@@ -49,7 +52,7 @@ dag = DAG(
 
 
 def stellar_etl_internal_task(
-    dag, task_name, command, cmd_args=[], resource_cfg="default"
+    dag, task_name, command, cmd_args=[], resource_cfg="default", output_file=""
 ):
     namespace = conf.get("kubernetes", "NAMESPACE")
 
@@ -68,6 +71,9 @@ def stellar_etl_internal_task(
 
     image = "{{ var.value.stellar_etl_internal_image_name }}"
 
+    etl_cmd_string = " ".join(cmd_args)
+    arguments = f""" {command} {etl_cmd_string} 1>> stdout.out 2>> stderr.out && cat stdout.out && cat stderr.out && echo "{{\\"output\\": \\"{output_file}\\"}}" >> /airflow/xcom/return.json"""
+
     return KubernetesPodOperator(
         task_id=task_name,
         name=task_name,
@@ -79,8 +85,9 @@ def stellar_etl_internal_task(
             "RETOOL_API_KEY": "{{ var.value.retool_api_key }}",
         },
         image=image,
-        cmds=[command],
-        arguments=cmd_args,
+        cmds=["bash", "-c"],
+        arguments=[arguments],
+        do_xcom_push=True,
         dag=dag,
         is_delete_operator_pod=True,
         startup_timeout_seconds=720,
@@ -104,7 +111,6 @@ retool_filepath = os.path.join(
     "retool-exported-entity.txt",
 )
 
-
 retool_export_task = stellar_etl_internal_task(
     dag,
     "export_retool_data",
@@ -121,6 +127,7 @@ retool_export_task = stellar_etl_internal_task(
         "--output",
         retool_filepath,
     ],
+    output_file=retool_filepath,
 )
 
 retool_table_name = "retool_entity_data"
@@ -138,7 +145,7 @@ retool_source_objects = [
     + retool_source_object_suffix
 ]
 
-task_vars = {
+retool_task_vars = {
     "task_id": f"del_ins_{retool_table_name}_task",
     "project": retool_public_project,
     "dataset": retool_public_dataset,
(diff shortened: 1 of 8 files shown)


schemas/retool_entity_data_schema.json
```diff
@@ -0,0 +1,163 @@
+[
```
amishas157, line 1: Took this JSON schema from bigquery after creating the table

dags/external_data_dag.py
```diff
@@ -0,0 +1,123 @@
...
+    command="export-retool",
+    cmd_args=[
+        "--start-time",
+        "{{ subtract_data_interval(dag, data_interval_start).isoformat() }}",
+        "--end-time",
+        "{{ subtract_data_interval(dag, data_interval_end).isoformat() }}",
+    ],
+    use_gcs=True,
+    env_vars={
+        "RETOOL_API_KEY": access_secret("retool-api-key", "default"),
```
amishas157, line 70: Saved the key in kubernetes cluster

dags/external_data_dag.py
```diff
@@ -0,0 +1,123 @@
...
+EXTERNAL_DATA_PROJECT_NAME = Variable.get("bq_project")
+EXTERNAL_DATA_DATASET_NAME = Variable.get("bq_dataset")
+RETOOL_TABLE_NAME = EXTERNAL_DATA_TABLE_NAMES["retool_entity_data"]
+RETOOL_EXPORT_TASK_ID = "export_retool_data"
+
+# Initialize the DAG
+dag = DAG(
+    "external_data_dag",
+    default_args=get_default_dag_args(),
+    start_date=datetime(2024, 12, 5, 14, 30),
```
chowbao, line 39: I recommend updating this start date. Generally because this is set to run `"0 22 * * *"` you don't want to set the hour::minutes part of the datetime

dags/external_data_dag.py
```diff
@@ -0,0 +1,123 @@
...
+    description="This DAG exports data from external sources such as retool.",
+    schedule_interval="0 22 * * *",
+    params={
+        "alias": "external",
```
chowbao, line 43: Can you remind me what this alias does again?

dags/external_data_dag.py
chowbao: Outside of this PR do you mind adding the api key to [hubble GCP secret manager](https://console.cloud.google.com/security/secret-manager?referrer=search&inv=1&invt=AbkT7g&project=hubble-261722)?

dags/external_data_dag.py
amishas157: Used for generating batch id

https://github.com/stellar/stellar-etl-airflow/blob/042908ceefd562ae29843a08914c9c2c026de377/dags/stellar_etl_airflow/macros.py#L10-L12

dags/external_data_dag.py
```diff
@@ -0,0 +1,123 @@
...
+        "{{ subtract_data_interval(dag, data_interval_end).isoformat() }}",
+    ],
+    use_gcs=True,
+    env_vars={
+        "RETOOL_API_KEY": access_secret("retool-api-key", "default"),
+    },
+)
+
+
+def get_insert_to_bq_task(
```
chowbao, line 75: Is this function different than https://github.com/stellar/stellar-etl-airflow/blob/master/dags/stellar_etl_airflow/build_bq_insert_job_task.py or anything else in `./dags/stellar_etl_ariflow/build_*`?

If it's different it might be worthwhile to separate this out into either `utils.py` or another `*.py file`

dags/external_data_dag.py
amishas157: Added the key to GCP secret manager as well

dags/external_data_dag.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+retool_insert_to_bq_task = get_insert_to_bq_task(
+    table_name=RETOOL_TABLE_NAME,
+    project=EXTERNAL_DATA_PROJECT_NAME,
+    dataset=EXTERNAL_DATA_DATASET_NAME,
+    export_task_id=RETOOL_EXPORT_TASK_ID,
+    source_object_suffix="",
+    partition=False,
+    cluster=False,
```
chowbao, line 119: Did you set these to false because the data is so small?

amishas157: Yep, exactly

dags/stellar_etl_airflow/build_internal_export_task.py
```diff
@@ -0,0 +1,108 @@
...
+    requests = {
+        "cpu": f"{{{{ var.json.resources.{resource_cfg}.requests.cpu }}}}",
+        "memory": f"{{{{ var.json.resources.{resource_cfg}.requests.memory }}}}",
+    }
+    container_resources = k8s.V1ResourceRequirements(requests=requests)
+
+    image = "{{ var.value.stellar_etl_internal_image_name }}"
+
+    output_filepath = ""
+    if use_gcs:
```
chowbao, line 53: nit: technically the `gcs` if/case could be split out into a separate function or interface so that it's easy to add support for other cloud storage systems. BUT I don't think we'll ever need support outside of GCS for this so it's fine to leave this as is

chowbao (approved):
LGTM left some minor comments

dags/external_data_dag.py
amishas157: I think, it is different since it is combination of delete and insert. Whereas the one in `build_bq_insert_job_task` is just insertion

dags/stellar_etl_airflow/build_internal_export_task.py
amishas157: Agree. I think in general we could try to write the operators as classes, which can help with overriding.. Right now, it is more functional.

dags/external_data_dag.py
chowbao: Ah that's fine then. Yeah a nit to move to a separate helper file. Also recommend renaming the function to be more descriptive of it deleting and inserting instead of just `get_insert_to_bq_task`. Maybe just add `delete` to the function name

dags/stellar_etl_airflow/build_internal_export_task.py
chowbao: Honestly one day we'll probably refactor all of stellar-etl-airflow

dags/external_data_dag.py
amishas157: So I moved it to `dags/stellar_etl_airflow/build_del_ins_operator.py` in https://github.com/stellar/stellar-etl-airflow/pull/550/commits/9a4e60f377785229e35000eb4f3917f222fad1ed

Also renamed the method to `create_export_del_insert_operator`. Essentially `create_export_del_insert_operator` is specific type of `create_del_ins_task`

dags/stellar_etl_airflow/build_internal_export_task.py
amishas157: 🤞

dags/external_data_dag.py
amishas157: Done in https://github.com/stellar/stellar-etl-airflow/pull/550/commits/9a4e60f377785229e35000eb4f3917f222fad1ed

PR #552, opened 2024-12-19
Update sandbox dag with retool

chowbao:
Fix sandbox_update_dag to pull retool data from the correct project.dataset

### DIFF
--- a/dags/sandbox_update_dag.py
+++ b/dags/sandbox_update_dag.py
@@ -45,7 +45,6 @@ with DAG(
     BQ_DATASET = Variable.get("public_dataset")
     SANDBOX_PROJECT = Variable.get("bq_project")
     SANDBOX_DATASET = Variable.get("sandbox_dataset")
-    INTERNAL_DATASET = Variable.get("bq_dataset")
 
     batch_run_date = "{{ batch_run_date_as_datetime_string(dag, data_interval_start) }}"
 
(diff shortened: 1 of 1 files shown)


dags/sandbox_update_dag.py
```diff
@@ -71,6 +70,11 @@
             "target_dataset": SANDBOX_DATASET,
             "batch_run_date": batch_run_date,
         }
+
+        if table_id == "retool_entity_data":
+            sql_params["project_id"] = SANDBOX_PROJECT
+            sql_params["dataset_id"] = INTERNAL_DATASET
+
```
sydneynotthecity, line 77: Do we need to copy this table over to the sandbox? I _believe_ canvas has full access to `crypto_stellar_internal_2` tables, but we may need to confirm. Since the table is only 30 rows, there is no reason for us to rebuild

chowbao: Oh I didn't know that. If they have access then I can change this to just exclude retool_entity_data instead

sydneynotthecity: Update: canvas does not have access to the dataset. My preference would be to build a view in the sandbox instead of a table

chowbao: Updated to skip `retool_entity_data`
312cd07001fdf2d66dce4fd507269d64261572d6

chowbao: View created in sandbox

sydneynotthecity (approved):
LGTM

PR #574, opened 2025-03-31
Add token transfer dbt task

sydneynotthecity:
<details>

### PR Structure

      otherwise).

### Thoroughness

</details>

### What

Minor Release:
Releasing the token transfers table, a new table that tracks value sent on the network. This table includes both classic and soroban events. For more information, see: https://github.com/stellar/stellar-dbt/pull/318

Also fixed a bug in the `singular_tests` DAG, where the DAG retries 5+ times. The DAG is an alert mechanism for data quality issues. If a test fails, that is a true failure. There is no need to retry multiple times because it triggers multiple alerts in Slack, making the alerts noisy.

### Known limitations

This is a pre-protocol 23 version of events, meaning the events are derived from the raw rxmeta instead of produced by core itself. This table will be refactored to use core events as its source of record once protocol 23 is released.

### DIFF
--- a/airflow_variables_dev.json
+++ b/airflow_variables_dev.json
@@ -372,6 +372,7 @@
     "snapshot_state": 600,
     "soroban": 720,
     "trade_agg": 720,
+    "token_transfer": 1200,
     "update_sandbox": 60,
     "wallet_metrics": 720
   },
--- a/airflow_variables_dev.json
+++ b/airflow_variables_dev.json
@@ -124,7 +124,7 @@
     "partnership_assets__account_holders_activity_fact": false,
     "partnership_assets__asset_activity_fact": false
   },
-  "dbt_image_name": "stellar/stellar-dbt-dev:90d2368f9",
+  "dbt_image_name": "stellar/stellar-dbt-dev:b459a809b",
   "dbt_internal_source_db": "test-hubble-319619",
   "dbt_internal_source_schema": "test_crypto_stellar_internal",
   "dbt_job_execution_timeout_seconds": 300,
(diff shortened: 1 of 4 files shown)


airflow_variables_dev.json
```diff
@@ -124,7 +124,7 @@
     "partnership_assets__account_holders_activity_fact": false,
     "partnership_assets__asset_activity_fact": false
   },
-  "dbt_image_name": "stellar/stellar-dbt:b655e9f6c",
+  "dbt_image_name": "stellar/stellar-dbt-dev:90d2368f9",
```
sydneynotthecity, line 127: will update once testing is finished

dags/dbt_singular_tests_dag.py
```diff
@@ -17,6 +17,7 @@
         "container_resources": lambda s: k8s.V1ResourceRequirements(requests=s),
     },
     max_active_runs=1,
+    retries=0,  # DAG is an alerting mechanism for data quality issues, so no retries
```
chowbao, line 20: nit: I'm not sure if this should be zero. Technically we'd want it to retry on transient errors. But it reruns every 30 mins anyways so it should be okay

chowbao (approved):

amishas157 (approved):

dags/dbt_singular_tests_dag.py
sydneynotthecity: I upped to 1. There is supposed to be a [check](https://github.com/stellar/stellar-etl-airflow/blob/master/dags/stellar_etl_airflow/utils.py#L26) in airflow that retries specifically for transient errors and skips dbt errors, but i don't think it's working 🤷‍♀️

PR #577, opened 2025-04-07
Add DAG to create snapshot

amishas157:
<details>

### PR Structure

      otherwise).

### Thoroughness

</details>

### What

[TODO: Short statement about what is changing.]

### Why

[TODO: Why this change is being made. Include any context required to understand the why.]

### Known limitations

[TODO or N/A]


dags/dag_snapshot.py
```diff
@@ -0,0 +1,60 @@
...
+dag = DAG(
+    "dbt_snapshot",
+    default_args=get_default_dag_args(),
+    start_date=datetime(2025, 3, 20, 0, 1),
+    description="This DAG runs dbt models at a daily cadence",
+    schedule_interval="0 1 * * *",  # Runs at 01:00 UTC
+    user_defined_filters={
+        "container_resources": lambda s: k8s.V1ResourceRequirements(requests=s),
+    },
+    max_active_runs=3,
```
sydneynotthecity, line 25: This will need to be updated to 1 so that tasks don't accidentally run in parallel

sydneynotthecity: If you want to parallelize tasks, the param is `max_active_tasks`

dags/dag_snapshot.py
```diff
@@ -0,0 +1,60 @@
...
+    # sla_miss_callback=alert_sla_miss,
+)
+
+trustline_snapshot_task = dbt_task(
+    dag,
+    tag="custom_snapshot_trustline",
+    excluded="stellar_dbt_public",
+    env_vars={
+        "BACKFILL_START_DATE": "{{ ds if run_id.startswith('scheduled_') else params.backfill_start_date }}",
+        "BACKFILL_END_DATE": "{{ macros.ds_add(ds, 1) if run_id.startswith('scheduled_') else params.backfill_end_date }}",
```
sydneynotthecity, line 45: in case we want to adjust the batch interval, it's more precise to use `next_ds` which is the next batch date: https://airflow.apache.org/docs/apache-airflow/1.10.13/macros-ref.html

dags/dag_snapshot.py
sydneynotthecity, line 1: there is a dag config option, `depends_on_past` that we can set to `True` so that dag runs are dependent on prior dag runs

sydneynotthecity (commented):
I know this is still a WIP, just wanted you aware of some airflow config options that we should use

dags/dag_snapshot.py
amishas157: ✨

dags/dag_snapshot.py
amishas157: Yep, that would be perfect

dags/dag_snapshot.py
```diff
@@ -0,0 +1,77 @@
...
+    get_default_dag_args,
+    init_sentry,
+)
+
+init_sentry()
+
+dag = DAG(
+    "dbt_snapshot",
+    default_args={**get_default_dag_args(), **{"depends_on_past": True}},
+    start_date=datetime(2025, 3, 20, 0, 1),
```
chowbao, line 19: Is there a reason to start this at the 1 second?

dags/dag_snapshot.py
```diff
@@ -0,0 +1,77 @@
...
+    env_vars={
+        "SNAPSHOT_START_DATE": "{{ ds if run_id.startswith('scheduled_') else params.snapshot_start_date }}",
+        "SNAPSHOT_END_DATE": "{{ next_ds if run_id.startswith('scheduled_') else params.snapshot_end_date }}",
+        "SNAPSHOT_FULL_REFRESH": "{{ false if run_id.startswith('scheduled_') else params.snapshot_full_refresh }}",
+    },
+)
+
+trustline_snapshot_task
+accounts_snapshot_task
+claimable_balances_snapshot_task
```
chowbao, line 77: Do you think we need to have a trigger/watch on the state table dag runs completing? Kind of like the dbt eho dag?

amishas157: So this is dependent on staging tables, we could make it wait on success of enriched_base _tables dag scheduled for 11:30 utc ?

chowbao: Is that 11:30 PM UTC? Then yeah that sounds good

chowbao (approved):
LGTM. Merge after addressing comments

PR #584, opened 2025-04-16
Add date_macro param to dbt tasks

sydneynotthecity:
<details>

### PR Structure

      otherwise).

### Thoroughness

</details>

### What

Add a new parameter to `dbt_task`, `date_macro` which allows the user to specific which [airflow macro](https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html) to pass as the execution date. This gives developers the flexibility to date parameterize their dbt code by date, `YYYY-MM-DD` or timestamp.

### Why

The incremental strategy insert_overwrite partitions executes its dbt model in multiple steps. There is an end hook that is a merge statement that passes the execution date as the partition to replace. Because Airflow was passing a timestamp, this code could not replace date partitions. This keeps consistency in the code without having to change the date logic in dbt in multiple places

### Known limitations

[TODO or N/A]

### DIFF
--- a/dags/stellar_etl_airflow/build_dbt_task.py
+++ b/dags/stellar_etl_airflow/build_dbt_task.py
@@ -8,6 +8,13 @@ from kubernetes.client import models as k8s
 from stellar_etl_airflow.default import alert_after_max_retries
 from stellar_etl_airflow.utils import skip_retry_dbt_errors
 
+# Define valid Airflow date macros
+VALID_DATE_MACROS = {
+    "ts": "{{ ts }}",
+    "ds": "{{ ds }}",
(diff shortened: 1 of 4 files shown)


dags/stellar_etl_airflow/build_dbt_task.py
```diff
@@ -67,7 +74,28 @@ def dbt_task(
     resource_cfg="default",
     run_singular_test="false",
     run_recency_test="false",
+    date_macro="ts",
```
sydneynotthecity, line 77: this was the old hardcoded value. setting as the default so that no existing DAG definitions break

chowbao (approved):

